PR #1, opened 2023-01-22
Setup Thicket for PyPI hosting and Pip Installation

ilumsden:
This PR will tweak the build, setup, and configuration files for Thicket so that Thicket will be able to be uploaded to PyPI and installed using `pip`.

Currently, this PR does the following:
* Adds build system info to `pyproject.toml` (see [PEP 517](https://peps.python.org/pep-0517/) for more info). This is the recommended way to configure a modern Python project for building. Additionally, in the (possibly not-too-distant) future, this will be required for any `pip`-installable package.
* Adds a `setup.cfg` file to configure `setuptools` when running `pip` in its PEP 517-compliant mode. This can be fully migrated into `pyproject.toml` once certain features of `setuptools` move out of beta.
* Adds a `python_requires` argument in `setup.py` since Hatchet (and, by extension, Thicket) only support Python 2.7 and Python 3.5+.

Additional information and details will be added through comments on this PR as they become available.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ setup(
     long_description_content_type="text/markdown",
     keywords="",
     url="",
-    python_requires = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*",
+    python_requires=">=3.6.1",
     packages=[
         "thicket",
         "thicket.stats",
@@ -42,7 +42,7 @@ setup(
         "pydot",
         "matplotlib",
         "numpy",
-        "pandas",
+        "pandas >= 1.1",
         "llnl-hatchet",
     ],
 )
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ setup(
     long_description_content_type="text/markdown",
     keywords="",
     url="",
+    python_requires = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*",
     packages=[
         "thicket",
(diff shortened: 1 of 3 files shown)


slabasan (changes requested):
Can you fix the black errors and then we can merge this?

setup.py
```diff
@@ -42,5 +43,6 @@ def readme():
         "matplotlib",
         "numpy",
         "pandas",
+        "llnl-hatchet",
```
slabasan, line 46: I will need to update the tag of llnl-hatchet for this to work. We had been using the tip of develop as the llnl-hatchet dependency during our development, but we can tag a new hatchet.

setup.py
```diff
@@ -30,6 +30,7 @@ def readme():
     long_description_content_type="text/markdown",
     keywords="",
     url="",
+    python_requires = ">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*",
```
slabasan: This list of python versions might be more narrow. Thicket is dependent on some newer features in pandas and numpy that are only available in python3 and later. Need to verify what those versions are to determine this final list.

ilumsden: I went through Thicket's use of pandas and found that Thicket requires pandas 1.1 or higher. I have adjusted the pandas entry in `install_requires` accordingly.

Based on the pandas documentation, pandas 1.1 only officially supports Python 3.6.1 or higher. So, I have adjusted `python_requires` to be `>=3.6.1`.

PR #2, opened 2023-01-23
Adds new GitHub Action to publish to PyPI on GitHub Release

ilumsden:
This PR adds a new GitHub Action that will publish a new Thicket release on PyPI whenever a new GitHub release is published.

For this Action to work, a PyPI API token will have to be added as a secret to this repo. This article explains the process of obtaining the API token and adding it to the repo: https://www.seanh.cc/2022/05/21/publishing-python-packages-from-github-actions/#create-a-pypi-api-token

The name of the token should be `PYPI_API_TOKEN`. If it's not, I'll have to change the last line of the `.yaml` file for the Action refer to the correct secret.

@slabasan, if we want to add this Action, it would probably be best if you added the secret.

### DIFF
--- a/.github/workflows/pypi_build_and_upload.yaml
+++ b/.github/workflows/pypi_build_and_upload.yaml
@@ -1,0 +1,26 @@
+name: Build Thicket and Upload to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build_and_upload:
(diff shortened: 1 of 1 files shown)


slabasan:
@ilumsden I've created and added the secret to this project

ilumsden:
Replaced by #27

PR #3, opened 2023-01-24
Various setuptools fixes to prepare for PyPI release

ilumsden:
There are several tweaks that have to be made to get Thicket ready for its first release on PyPI. This PR makes those changes.

Comments will be added to this PR as new changes are introduced.

Currently, this PR does the following:
* Adds `wheel` as a build system dependency in `pyproject.toml` (needed to build bdist objects through a PEP 517 frontend)
* Adds a `MANIFEST.in` file to make sure all the JavaScript and HTML code for `thicket.vis` is correctly added to the sdist and bdist objects
* ~~Adds a custom setuptools `build_py` step that builds the non-Python code in `thicket.vis` using NPM~~
* Adds an extra in `setup.py` that will allow you to install Thicket with an Extra-P dependency by using `pip install llnl-thicket[extrap]`


ilumsden:
Additional changes:
* Adds a new step to the unit testing GitHub Action that will install NPM
  * _Note: this may be removed if we decide to just produce a warning if NPM is not found._
* Updates the README to include instructions about pip installation and building the interactive visualizations
* Removes `setup.cfg` for the time being to maximize backwards compatibility with old (mainly Python 2.7 focused) versions of `setuptools`. All the information in `setup.cfg` is preserved in `setup.py`

ilumsden:
There's been a slight change to the way that the HTML and JavaScript in `thicket.vis` is being built. To make sure the building will work for wheel-based installs too, the HTML/JavaScript building will now be done on-the-fly by code in `thicket/vis/__init__.py`. As a result, the first time you access the `thicket.vis` module, NPM will be run to build the HTML and JavaScript code. If NPM is not installed when `thicket.vis` is accessed, a `FileNotFoundError` will be raised.

slabasan (approved):

PR #4, opened 2023-01-24
 Removed vega boxplots import and cleaned up the code a bit 

cscully-allison:
Changes:

- The topdown vis call from python only now requires the thicket object argument, use of `node` and `test` args have been removed.


ilumsden (approved):
This all looks good to me

PR #6, opened 2023-03-16
use manual codecov upload

slabasan:


ilumsden:
We've got an automated integration of Codecov into our CI now, so this is no longer needed.

PR #8, opened 2023-04-21
Add and Refactor Unit Tests

michaelmckinsey1:
- Remove redundant data files and their corresponding fixtures and references, `data/misc`.
- Separate `filter` unit tests into their own file.
- Add `test_thicketize_graphframe` and `test_unify_ensemble`.
- Update & add to existing unit tests where necessary.
- Override `__eq__` operator in Thicket.


slabasan (changes requested):
Let's rename test_readers.py with test_caliperreader.py. In the future when we have other formats, it'll be easier for us to keep them in separate files.

slabasan (approved):

PR #17, opened 2023-04-28
Make Thicket Construction Print Statements Optional

michaelmckinsey1:
Referring to these:
```
Union Graph == thicket[1].graph
Union Graph == thicket[1].graph
Union Graph == thicket[1].graph
```
With large datasets, this becomes an annoyance, see https://lc.llnl.gov/gitlab/pave/thicket/-/issues/23

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -683,24 +683,41 @@ class Thicket(GraphFrame):
 
         return union_graph
 
-    def unify_pairwise(th_list):
+    @staticmethod
+    def unify_pairwise(th_list, debug=False):
         """Unifies two thickets graphs and dataframes.
 
         Ensure self and other have the same graph and same node IDs. This may
         change the node IDs in the dataframe.
 
         Update the graphs in the graphframe if they differ.
+
+        Arguments:
+            th_list (list): list of Thicket objects
+            debug (bool): print debug statements
+
+        Returns:
+            union_graph (Graph): unified graph
         """
         union_graph = th_list[0].graph
         for i in range(len(th_list)):
             for j in range(i + 1, len(th_list)):
-                print("Unifying (" + str(i) + ", " + str(j) + "...")
+                if debug:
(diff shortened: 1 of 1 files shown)


thicket/thicket.py
```diff
@@ -683,24 +683,41 @@ def unify_pair(self, other):
 
         return union_graph
 
-    def unify_pairwise(th_list):
+    @staticmethod
```
slabasan, line 686: Do these functions need to be static? If not, we should make another PR to remove this specification.

thicket/thicket.py
```diff
@@ -683,24 +683,41 @@ def unify_pair(self, other):
...
+    def unify_pairwise(th_list, debug=False):
```
slabasan, line 687: If we are going to have more debug statements throughout thicket, we might think about having a global enable/disable debug mode that we can specify from an environment variable or config file.

slabasan (commented):
LGTM, just 1 question about staticmethods.

thicket/thicket.py
michaelmckinsey1: My current strategy is to use this decorator whenever we are not using `self` inside a class function. Otherwise, you will run into [errors](https://stackoverflow.com/a/43587154) when trying to call the function outside of the class.

I suppose we could remove them from any methods in which we don't want any users to run. But then would we want to prepend nothing, `_`, or `__` before these functions?

thicket/thicket.py
michaelmckinsey1: If we want to use [dotenv](https://pypi.org/project/python-dotenv/) I can add that here. I don't think any other functions are using a debug variable at the moment.

slabasan: We don't need to implement it now, just something for us to keep in mind if we find we want to have more debugging capabilities.

thicket/thicket.py
slabasan: Got it, makes sense.

PR #25, opened 2023-05-22
Update filter_metadata() test case for two columns

vanessalama09:
The goal of this PR is to add to the filter unit test for cases where user inputs two columns and conditions for filtering the metadata.


slabasan (commented):
What does the total_col= parameter specify? Are we testing for 1 match vs 2 matches (AND, OR)? If so, we should split these into two separate functions and update the name of the functions.

slabasan (approved):

PR #27, opened 2023-05-24
Adds GitHub Action to build and upload wheels and sdist

ilumsden:
This PR ports Hatchet's new GitHub build-and-upload Action to Thicket. When triggered, this Action will build a sdist and wheels for Thicket. If the trigger is a tag push, the Action will also upload the build artifacts to TestPyPI. On the other hand, if the trigger is a release publication, the Action will upload the artifacts to regular PyPI.

### DIFF
--- a/.github/workflows/build_and_upload_wheels.yaml
+++ b/.github/workflows/build_and_upload_wheels.yaml
@@ -1,0 +1,160 @@
+name: Build wheels
+
+on:
+  Uncomment for testing through a PR
+  pull_request:
+    branches: [develop, releases/**]
+  workflow_dispatch:
+  release:
+    types:
+      - published
+  push:
+    tags:
+      - 'v**'
+
+jobs:
+
+  build_wheels_linux_3:
+    name: Build wheels for Linux
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up QEMU to support non-x86 architectures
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_upload_wheels.yaml
```diff
@@ -0,0 +1,191 @@
...
+jobs:
+
+  # TODO: if we ever add compiled code to Thicket (e.g., Cython modules),
+  #       remove the build_wheels job
+  build_wheels:
+    name: Build wheels for Linux
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9, '3.10', 3.11]
```
slabasan, line 24: Why is python 3.10 in quotes here?

ilumsden: It's a limitation/feature of YAML (depends on your perspective). Without quotes, each of those values are treated as floating point. So, when the runner sees `3.10`, it interprets it as `3.1`. By putting `3.10` in quotes, it is interpreted as a string, so the runner will see the correct value. This isn't an issue with any of the other values, so I didn't quote them.

slabasan: I see, thanks for clarifying. That is a nuanced behavior for sure!

slabasan (approved):

PR #28, opened 2023-05-25
Updated visualization functions

Treece-Burgess:
There are three visualization functions that have been pushed up for this PR:

- `display_boxplot`
- `display_heatmap`
- `display_historgram`

Each of the above listed visualization functions have now added the ability to work with a single indexed thicket or a multi indexed thicket (columnar_join).

### DIFF
--- a/thicket/stats/display_boxplot.py
+++ b/thicket/stats/display_boxplot.py
@@ -72,7 +72,7 @@ def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
         for i in range(1, len(columns)):
             if initial_idx != columns[i][0]:
                 raise ValueError(
-                    "Tuples must have the same column index for a columnar joined thicket."
+                    "Columns specified as tuples must have the same column index (first element)"
                 )
             else:
                 cols.append(columns[i][1])
--- a/thicket/stats/display_boxplot.py
+++ b/thicket/stats/display_boxplot.py
@@ -5,25 +5,33 @@
 
 import pandas as pd
 import seaborn as sns
+
 from ..utils import verify_thicket_structures
 
 
 def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
-    """Display a boxplot for each user passed node(s) and column(s). The passed nodes and columns must be from the performance data table.
+    """Display a boxplot for each user passed node(s) and column(s). The passed nodes
+    and columns must be from the performance data table.
 
-    Designed to take in a thicket, and display a plot with one or more boxplots depending on the number
-    of nodes and columns passed.
+    Designed to take in a thicket, and display a plot with one or more boxplots
+    depending on the number of nodes and columns passed.
 
     Arguments:
-        thicket (thicket) : Thicket object
-        nodes (list) : List of nodes to view on the x-axis
-        column (list) : List of hardware/timing metrics to view on the y-axis.
-                        Note, if using a columnar_joined thicket a list of tuples must be
-                        passed in with the format: (column index, column name).
+        thicket (thicket): Thicket object
+        nodes (list): List of nodes to view on the x-axis
+        column (list): List of hardware/timing metrics to view on the y-axis.
+                        Note, if using a columnar_joined thicket a list of tuples must
+                        be passed in with the format: (column index, column name).
 
     Returns:
         (matplotlib Axes): Object for managing boxplot.
     """
+    for node in nodes:
+        if str(type(node)) != "<class 'hatchet.node.Node'>":
+            raise ValueError(
+                "Value(s) passed to node argument must be of type hatchet.node.Node."
+            )
+
     verify_thicket_structures(
         thicket.dataframe, index=["node", "profile"], columns=columns
     )
@@ -53,14 +61,13 @@ def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
             )
         else:
             return sns.boxplot(data=filtered_df, x="node", y=" ", **kwargs)
-
     else:
         initial_idx, initial_col = columns[0][0], columns[0][1]
         cols = [initial_col]
-        for element in columns[1 : len(columns)]:
+        for element in columns[1:len(columns)]:
             if initial_idx != element[0]:
                 raise ValueError(
-                    "The columns argument must have the same column index throughout for a columnar joined thicket."
+                    "Tuples must have the same column index for a columnar joined thicket."
                 )
             else:
                 cols.append(element[1])
--- a/thicket/stats/display_boxplot.py
+++ b/thicket/stats/display_boxplot.py
@@ -53,7 +53,7 @@ def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
             for pos in idx:
                 position.append(pos)
 
-        # rename columns such that the x-axis label is "name" and not "node", tick marks
+        # rename columns such that the x-axis label is "node" and not "name", tick marks
         # will be node names
         filtered_df = df.loc[position].rename(
             columns={"node": "hatchet node", "name": "node"}
@@ -94,7 +94,7 @@ def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
             for pos in idx:
                 position.append(pos)
 
-        # rename columns such that the x-axis label is "name" and not "node", tick marks
+        # rename columns such that the x-axis label is "node" and not "name", tick marks
         # will be node names
         filtered_df = df.loc[position].rename(
             columns={"node": "hatchet node", "name": "node"}
--- a/thicket/stats/display_boxplot.py
+++ b/thicket/stats/display_boxplot.py
@@ -18,27 +18,68 @@ def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
     Returns:
         plot: Returns a boxplot
     """
-    df = pd.melt(
-        thicket.dataframe.reset_index(),
-        id_vars=["node", "name"],
-        value_vars=columns,
-        var_name="Performance counter",
-        value_name=" ",
-    )
 
-    position = []
-    for node in nodes:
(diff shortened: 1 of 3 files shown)


thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
 
 import pandas as pd
 import seaborn as sns
+from ..utils import verify_thicket_structures
```
slabasan, line 10: New line between pandas and seaborn imports and utils import (separate third party from local imports per pep8)

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
 
 
 def display_boxplot(thicket, nodes=[], columns=[], **kwargs):
-    """Designed to take in a Thicket, and plot a boxplot based on user passed node's and a single metric.
+    """Display a boxplot for each user passed node(s) and column(s). The passed nodes and columns must be from the performance data table.
```
slabasan: Can you double check the character width of your docstring? We're trying to keep them to 88 chars.

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
+
+    Designed to take in a thicket, and display a plot with one or more boxplots depending on the number
+    of nodes and columns passed.
 
     Arguments:
         thicket (thicket) : Thicket object
         nodes (list) : List of nodes to view on the x-axis
-        column (str) : A single hardware or timing metric to view on the y-axis
+        column (list) : List of hardware/timing metrics to view on the y-axis.
```
slabasan: Remove space before : in docstrings

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
+            columns={"node": "hatchet node", "name": "node"}
         )
+
+        if len(columns) > 1:
+            return sns.boxplot(
+                data=filtered_df, x="node", y=" ", hue="Performance counter", **kwargs
+            )
+        else:
+            return sns.boxplot(data=filtered_df, x="node", y=" ", **kwargs)
+
```
slabasan: Remove new line

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
     else:
-        return sns.boxplot(data=filtered_df, x="node", y=" ", **kwargs)
+        initial_idx, initial_col = columns[0][0], columns[0][1]
+        cols = [initial_col]
+        for element in columns[1 : len(columns)]:
```
slabasan: Let's remove spaces so it reads `[1 : len(columns)]`. I think this will still pass formatting checks.

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
+            if initial_idx != element[0]:
+                raise ValueError(
+                    "The columns argument must have the same column index throughout for a columnar joined thicket."
```
slabasan: This is the error you're displaying if the user passes 2 or more columns that are specified as a tuple, where the first element in the tuple is not the same, correct? Trying to think of a way to simplify this helper error message to limit how many times we say column in the same sentence.

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
+            value_name=" ",
+        )
+
+        position = []
+        for node in nodes:
+            idx = df.index[df["node"] == node]
+            for pos in idx:
+                position.append(pos)
+
+        filtered_df = df.loc[position].rename(
```
slabasan, line 99: Why do we need to rename columns here? When you're calling the boxplot functions, can you pass the original column names to the x= parameter?

thicket/stats/display_histogram.py
```diff
@@ -9,15 +9,17 @@
 
 
 def display_histogram(thicket, node=None, column=None, **kwargs):
-    """Display a histogram.
+    """Display a histogram for a user passed node and column. Node and column must come from the performance data table.
```
slabasan: Check width of chars in docstring.

thicket/stats/display_histogram.py
```diff
@@ -26,17 +28,33 @@ def display_histogram(thicket, node=None, column=None, **kwargs):
...
+            value_vars=column,
+            value_name=" ",
+        )
+
+        filtered_df = df[df["node"] == node]
+
+        ax = sns.displot(filtered_df, x=" ", kind="hist", **kwargs)
+
+        return ax
+
```
slabasan: Remove new line

thicket/stats/display_boxplot.py
```diff
@@ -5,40 +5,90 @@
...
     """
-    df = pd.melt(
-        thicket.dataframe.reset_index(),
-        id_vars=["node", "name"],
-        value_vars=columns,
-        var_name="Performance counter",
-        value_name=" ",
+    verify_thicket_structures(
+        thicket.dataframe, index=["node", "profile"], columns=columns
     )
```
slabasan, line 39: Should we add a check to make sure the list of nodes is of node type?

slabasan (changes requested):
One other question, how many datasets have you tested with this vis functionality? Is this code robust against the different datasets we have?

thicket/stats/display_boxplot.py
Treece-Burgess: Yes, that is correct. Possibly could use: Columns must have same index for a passed in columnar_joined thicket. If they are confused, they can run `help()` on the function they are using. Upon running `help()` they should be able to see the python doc string which displays the format needed if a user passes in a columnar_joined thicket.

thicket/stats/display_boxplot.py
Treece-Burgess: You are wanting to remove the spaces from: `[1  :  len(columns)]`? So it would look like the following: 
`[1 : len(columns)]`. Only issue with that could be that flake or black complains about it. The reason I have the spaces there in the first place is due to the formatting checks I ran the other day.

thicket/stats/display_boxplot.py
Treece-Burgess: Originally we output the hatchet node id (the dictionary) onto the x-axis, after a demo it was decided that instead of using the dictionary node id, we just output the name instead since it would not take up as much space. Without renaming the name column, the x-axis label would end up being `name` instead of `node`. Which is why I rename the columns here.

thicket/stats/display_boxplot.py
Treece-Burgess: Next push up will have the space between third party imports and local imports. I am going to go ahead and resolve this comment.

thicket/stats/display_boxplot.py
Treece-Burgess: I have double checked the sentence width with the python doc strings, and formatted such that we keep them to 88 chars. This will be shown in the next push up. Going to resolve this comment now.

thicket/stats/display_boxplot.py
Treece-Burgess: Removed space before `:` in docstring for `display_boxplot.py`. This will be seen within the next push up. Going to go ahead and resolve conversation.

thicket/stats/display_boxplot.py
Treece-Burgess: Removed new line at line 56 within `display_boxplot.py`. This will be seen in the next push up. Going to go ahead and resolve conversation.

thicket/stats/display_boxplot.py
Treece-Burgess: Black does complain if you remove the spaces, but I have gone ahead ahead and pushed it up to see if it passes the checks.

thicket/stats/display_boxplot.py
Treece-Burgess: I have added checks in `display_histogram.py` and `display_boxplot.py`.

thicket/stats/display_boxplot.py
```diff
@@ -6,39 +6,96 @@
...
-
-    position = []
     for node in nodes:
-        idx = df.index[df["node"] == node]
-        for pos in idx:
-            position.append(pos)
+        if str(type(node)) != "<class 'hatchet.node.Node'>":
+            raise ValueError(
+                "Value(s) passed to node argument must be of type hatchet.node.Node."
+            )
```
Treece-Burgess: @slabasan I have added node check here and it will be the same for `display_boxplot`. Llet me know if this looks fine or if you had envisioned something else.

thicket/stats/display_boxplot.py
Treece-Burgess: Seaborn uses the column header as the main x-axis label. As discussed previously we wanted to use the name column as the tick marker labels instead of the hatchet node dictionary. Therefore, if we pass `name` to perform the following code snippet: `sns.boxplot(data=filtered_df, x="name", y=" ", **kwargs)` we would end up having the main x-axis label as `name` due to how Seaborn works. 

Due to this I rename the `name` column to be `node` such that we get the shortened `name` tick labels and we get the main x-axis label to be `node` such that a user will know that they are looking at a `node`.

slabasan: See updated code, I've added a statement to indicate the need for the renaming. Let me know what you think.

Treece-Burgess: I believe that `name` and `node` need to be swapped within this sentence: _rename columns such that the x-axis label is "name" and not "node"_.

So it would read: _rename columns such that the x-axis label is "node" and not "name"_. If we didn't rename here then the x-axis label would end up being `name`.

slabasan (approved):

thicket/stats/display_boxplot.py
slabasan: Nice catch, made this change.

PR #44, opened 2023-06-09
Adds support for Hatchet's "new-style" queries to Thicket

ilumsden:
This PR adds support for Hatchet's new-style queries to Thicket. More specifically, this PR adds the following:
* A new `thicket.query` module that re-exports the Query Language classes and functions
* **Support for the object- and string-based dialects of the Query Language**

The dialect support is enabled through the new `multi_index_mode` parameter. This parameter is currently being documented in LLNL/hatchet#90.

### DIFF
--- a/thicket/query.py
+++ b/thicket/query.py
@@ -1,0 +1,60 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+from hatchet.query import (
+    Query,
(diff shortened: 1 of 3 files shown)


slabasan (approved):
Looks good, but I see it is still in draft mode. Is this ready for merge?

slabasan:
@ilumsden Let me know if this one is no longer in draft mode.

ilumsden:
@slabasan there are still some things I need to fix to get things passing with unit tests. I also realized that this will only get the dialects working for Thickets without a column multi-index. That means things like columnar join Thickets aren't going to be supported yet.

PR #48, opened 2023-06-16
Docs fixes

slabasan:
- [ ] fix fig 1 in user guide
- [ ] user guide call tree and performance data to match


vanessalama09 (commented):
Docs have all tutorials and demos. The gifs needed to be compressed to push to GitHub. Only remaining to replace the figures in User Guide.

PR #51, opened 2023-06-20
Preference Redesign

Treece-Burgess:
This PR is to address the redesign of the `StatsPreference` class which is currently located within `statspreference.py`. The goal is to migrate the `StatsPreference` class to a function. The `.py` file that houses the function is called `preference.py`. With the function taking on the following arguments:

-  thicket object: single-indexed or columnar joined thicket
- columns: list of exactly 2 columns must be passed
- comparison_func: lambda or python function a user can pass to use such that they can assign a preference
- test: Test to be performed


thicket/stats/preference.py
```diff
@@ -0,0 +1,111 @@
...
+                    comparison_func(
+                        thicket.statsframe.dataframe[columns[0] + "_std"][i],
+                        thicket.statsframe.dataframe[columns[1] + "_std"][i],
+                    )
+                )
+            else:
+                pref_mean.append("No preference")
+                pref_std.append("No preference")
+        thicket.statsframe.dataframe[
+            columns[0] + " vs " + columns[1] + "_std_preferred"
```
Julius-Plehn: Extract repeated string concatenation

thicket/stats/preference.py
```diff
@@ -0,0 +1,111 @@
...
+                    )
+                )
+            else:
+                pref_mean.append("No preference")
+                pref_std.append("No preference")
+        thicket.statsframe.dataframe[
+            (
+                "Preference",
+                str(columns[0]).replace("'", "")
+                + " vs "
```
Julius-Plehn: Extract repeated string concatenation

thicket/stats/ttest.py
```diff
@@ -0,0 +1,116 @@
...
+                equal_var=False,
+            )
+
+            t_statistics.append(tStatistic.statistic)
+        # store results into thicket's aggregated statistics table
+        thicket.statsframe.dataframe[
+            (
+                "Preference",
+                str(columns[0]).replace("'", "")
+                + " vs "
```
Julius-Plehn: Extract repeated string concatenation

thicket/stats/ttest.py
```diff
@@ -0,0 +1,116 @@
...
+                mean2=thicket.statsframe.dataframe[mean_columns[1]][i],
+                std2=thicket.statsframe.dataframe[std_columns[1]][i],
+                nobs2=nobs,
+                equal_var=False,
+            )
+
+            t_statistics.append(tStatistic.statistic)
+        # store results into thicket's aggregated statistics table
+        thicket.statsframe.dataframe[
+            columns[0] + " vs " + columns[1] + "_tvalue"
```
Julius-Plehn: Extract repeated string concatenation

thicket/stats/preference.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+    # thicket object wihtout columnar index
+    pref_mean = []
+    pref_std = []
+    if thicket.dataframe.columns.nlevels == 1:
+        for i, t_statistic in enumerate(t_statistics):
+            if t_statistic < -1 * tvalue or t_statistic > tvalue:
+                pref_mean.append(
+                    comparison_func(
+                        thicket.statsframe.dataframe[columns[0] + "_std"][i],
```
Julius-Plehn: Add mean instead of std

thicket/stats/ttest.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    # check to make sure alpha is passed
+    if "alpha" not in kwargs:
+        raise ValueError(
+            "For test=ttest, a user passed alpha value of 0 to 1 is needed."
+        )
+    else:
+        alpha = kwargs["alpha"]
+        # check to see if alpha value is between 0 and 1
+        if alpha < 0 or alpha > 1:
```
slabasan: Should this be an inclusive range? That is, can the alpha value be 1?

thicket/stats/ttest.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Returns:
+       tvalue (double): Value to be used to determine a preference within preference.py.
+       tstatistics (list): List of values to be used to determine a preference within
+            preference.py.
+    """
+    # check to make sure alpha is passed
+    if "alpha" not in kwargs:
+        raise ValueError(
+            "For test=ttest, a user passed alpha value of 0 to 1 is needed."
```
slabasan: Should we set a default alpha value? I recall in some code you've shown, the alpha value was set to 0.5? Is this for the same test?

thicket/stats/ttest.py
```diff
@@ -0,0 +1,122 @@
...
+        if alpha < 0 or alpha > 1:
+            raise ValueError("Value for alpha argument must be between 0 and 1.")
+
+    # check columns contain two columns
+    if len(columns) != 2:
+        raise ValueError("Columns must be a list of length 2.")
+
+    nobs = int(pd.unique(thicket.dataframe.groupby(level=0).size()))
+
+    # I believe profile numbers will be the same for single indexed thicket and columnar
```
slabasan: I'm not sure I understand this comment here. Is there an assumption about the profile hashes?

thicket/stats/ttest.py
```diff
@@ -0,0 +1,122 @@
...
+                str(columns[0]).replace("'", "")
+                + " vs "
+                + str(columns[1]).replace("'", "")
+                + "_tstatistic",
+            )
+        ] = t_statistics
+
+        return tvalue, t_statistics
+
+    else:
```
slabasan, line 92: Is this if/else for columnar indexed vs no columnar indexed thicket? If so, can you add comments to indicate this?

thicket/statspreference.py
```diff
@@ -1,249 +0,0 @@
-# Copyright 2022 Lawrence Livermore National Security, LLC and other
```
slabasan, line 1: I removed this file, my understanding is you have migrated this code to preference.py. Feel free to add back if this is not the case.

thicket/stats/preference.py
```diff
@@ -0,0 +1,114 @@
...
+from ..utils import verify_thicket_structures
+from .ttest import __ttest
+
+__statistical_tests = {"ttest": __ttest}
+
+
+def preference(thicket, columns, comparison_func, test="ttest", *args, **kwargs):
+    """Determine a preference between compilers, architecture, platform, etc.
+
+    Designed to take in a thicket and will append eight total columns to the
```
slabasan, line 15: What are the 8 columns that are added? I'm only seeing 5 columns, so not sure which ones I am missing.

thicket/stats/preference.py
```diff
@@ -0,0 +1,114 @@
...
+    aggregated statistics table. These columns will included: std, mean, tvalue,
+    tstatistic, and preferred. As a note, preferred will stand for the preferred
+    choice between two options.
+
+    Column names will have the following two formats:
+        1.
+          - column_name + _statistic where statistic could be either: std or mean
```
slabasan: Is the - to indicate a hyphen? So is the column name format <metric>-column_name_{mean/std}?

thicket/stats/preference.py
```diff
@@ -0,0 +1,114 @@
...
+        columns (list): List of hardware/timing metrics to determine a preference for.
+            Note, if using a columnar joined thicket a list of tuples must be passed in
+            with the format (column index, column name).
+        comparison_func (function): User-defined python or lambda function to decide a
+            preference.
+        test (str): User selected test.
+    """
+    if test not in __statistical_tests.keys():
+        raise ValueError("Test is not available.")
+    # want to discuss adding a more clear message for correct column format for columnar joined thicket
```
slabasan: Do we need to resolve this comment before we merge?

thicket/stats/preference.py
Treece-Burgess: The hyphen is suppose to be a representation of a list, and it would not be apart of the column name format. For example it suppose to be like: 

1. Begin numbered list <br>
    -  Hyphen or if you want to think of it as a bullet point you can as well

However, after looking after it using a hyphen does not make much sense actually so I will change it to:

Column names will have the following two formats:

1. column_name + _statistic where statistic could be either: std or mean
2. column_name + vs + column_name + _statistic where statistic could be
    either: tvalue, tstatistic, or preferred

thicket/stats/preference.py
Treece-Burgess: Originally within the `preference` function design I would check to make sure that the list of `columns` passed in for a columnar joined thicket would be in the format (column index, column name). However, I have removed that within the current redesign and currently rely on `verify_thicket_structures` to check. The error message however, isn't the most explicit. It will just state that columns don't exist within the thicket. Which for a single index thicket is fairly self explanatory, but since we add a layer of complexity with the multi-indexed/columnar joined thickets I believe adding a better error message would be a good idea. For example current error would look something like this:

```
th.preference(columnar_joined_th, columns=[column1, column2],...) # user calling the preference function

ValueError: Column's do not exist within thicket # output of verify thicket structure
```

I think a better avenue would be to do the following checks inside `verify_thicket_structures`:

1.  Check to see if it is a columnar joined thicket 
2.  If it is then make sure they are within the tuple format
     * If not then error out saying we need the format (column index, column name)
4. If they are within the tuple format does the tuple of column index and column name exist in the thicket
    * If not then error out saying either the column index, column name or both do not exist and suggest to run `get_perf_columns`
6. If everything checks out then continue with the preference function

Adding this functionality to `verify_thicket_structures` will also be useful to other functions within Thicket as well such as Connor's vis, helper functions, etc.

thicket/stats/preference.py
Treece-Burgess: Done will be shown in next push up.

thicket/stats/preference.py
Treece-Burgess: Done will be shown in the next push up.

thicket/stats/preference.py
Treece-Burgess: Done will be reflected in the next push up.

thicket/stats/ttest.py
Treece-Burgess: This needs to be changed to an inclusive range, I looked a bit more into it and from what I could find the alpha value needs to be between 0 and 1. You may be able to find extremely niche cases where you could set it to be say 0 or 1 (would have to actually write a proof for them), but I would avoid trying to find this and go along with what the rest of researchers do which is between 0 and 1. 

So the updated code will be:
```
if alpha <= 0 or alpha >= 1:
   raise ValueError("Value for alpha argument must be between 0 and 1.")
```

This will be reflected in the next push up.

thicket/stats/ttest.py
Treece-Burgess: Discussed during the UTK-LLNL meeting, and it was confirmed that the single-indexed thicket and columnar joined thicket will have the same profile numbers. 

Due to this discussion, it was also discussed to add more explicit naming convention for :
```
tvalue = t.ppf(
        q=1 - alpha / 2, df=nobs + nobs - 2
    ) 
```

Instead of `nobs + nobs` it will become `nobs_column1 + nobs_column2` with:

- nobs_column1: number of profiles/values for the first column passed by a user
- nobs_column2: number of profiles/values for the second column passed by a user

thicket/stats/ttest.py
Treece-Burgess: Yes it is, and I will add the comments to indicate. Will be reflected in the next push up.

thicket/stats/preference.py
Treece-Burgess: These will be the columns added: 

1. Column 1 mean
2. Column 1 std
3. Column 2 mean
4. Column 2 std
5. Column1 + Column2 tvalue
6. Column1 + Column2 tstatistic
7. Column1 + Column2 std preferred
8. Column1 + Column2 mean preferred

Let me know if you have any questions. There are 8 in total added since we call ttest.py within preference. However, if you want to split it up into columns for ttest in ttest and columns for preference in preference I can do that as well. Possibly the best course of action if more tests are added letter, because maybe we have more or less than 8 columns added.

thicket/stats/ttest.py
Treece-Burgess: Default alpha value will be added to the `ttest.py` file. This will be reflected within the next push up.

thicket/stats/ttest.py
Treece-Burgess: Done and will be reflected in the next push up.

thicket/stats/ttest.py
Treece-Burgess: Done and will be reflected in the next push up.

slabasan (approved):

PR #55, opened 2023-06-30
Draft: Fill Performance DataFrame

michaelmckinsey1:
# Summary
Current behavior dictates that a profile won't show up in the PerfData if it does not exist for that node. This causes bugs when operating on profiles as the tree is always full whereas a slice of a profile may not be. This PR forces the PerfData to always have a full structure, inserting `NaN`s where applicable.


michaelmckinsey1:
#57 New approach

PR #63, opened 2023-07-12
Intersection Bugfix

michaelmckinsey1:
# Summary
After #57, `intersection()` no longer returns the correct result. This PR simplifies the logic using query and fixes the bug.


michaelmckinsey1:
~Should pass once #64 merged~

slabasan (approved):

PR #67, opened 2023-07-13
StatsFrame Name Bugfix Level

michaelmckinsey1:
# Summary
When the statsframe columns are multiindex, the name column should reside at the same level as in the PerfData.

### DIFF
--- a/thicket/tests/test_thicket.py
+++ b/thicket/tests/test_thicket.py
@@ -71,7 +71,18 @@ def test_sync_nodes(example_cali):
     assert helpers._are_synced(th.graph, th.dataframe)
 
 
-def test_aggregated_statistics_table(example_cali):
+def test_statsframe(example_cali):
+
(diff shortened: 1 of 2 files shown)


thicket/tests/test_thicket.py
```diff
@@ -71,7 +71,18 @@ def test_sync_nodes(example_cali):
     assert helpers._are_synced(th.graph, th.dataframe)
 
 
-def test_aggregated_statistics_table(example_cali):
```
slabasan, line 74: Why the change of the test name here?

michaelmckinsey1: The pattern we follow with all the other test functions is to use the name of the attribute/function name. I think someone accidentally changed this function in a find and replace since it used to be `test_statsframe`

slabasan (approved):

PR #71, opened 2023-07-20
Relax Verify Thicket Structure Check for Stats Functions

Treece-Burgess:
This PR contains a single change which will be outlined below. 

1. Update to `verify_thicket_structure` check within stats functions
    - Removed `profile` from the list of indexes to check for, this allows users to join on any metadata column they would like now.

### DIFF
--- a/thicket/stats/calc_boxplot_statistics.py
+++ b/thicket/stats/calc_boxplot_statistics.py
@@ -9,7 +9,7 @@ import numpy as np
 from ..utils import verify_thicket_structures
 
 
-def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **kwargs):
+def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75]):
     """Calculate boxplots lowerfence, q1, q2, q3, iqr, upperfence, and outliers for each
     node in the performance data table.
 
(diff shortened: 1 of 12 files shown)


thicket/stats/calc_boxplot_statistics.py
```diff
@@ -9,7 +9,7 @@
 from ..utils import verify_thicket_structures
 
 
-def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **kwargs):
```
michaelmckinsey1: Should this be a separate PR?

Treece-Burgess: Originally this PR was for both the `**kwargs` and for relaxing the verify thicket structure checks. However, I changed it to just be for relaxing the verify thicket structure check. Line 12 that you are pointing out, was apart of an old push, but I just double checked the current files and I do not see the `**kwargs` within the list of arguments.

michaelmckinsey1: [Develop](https://github.com/LLNL/thicket/blob/develop/thicket/stats/calc_boxplot_statistics.py#L12) shows it in the current code and that's why it's in the diff. I think it's reasonable to have it in here if the title/description is updated or if it is separated out into a different PR; whichever is easier.

Treece-Burgess: I see what you mean now, I will wait to remove that with another PR when the **kwargs are addressed. Thanks for pointing that out to me.

michaelmckinsey1 (approved):

slabasan (approved):

PR #43, opened 2023-06-07
add tests for stats

slabasan:

PR #77, opened 2023-07-29
Add Pandas Groupby Functionality to Stats Functions

### LINKED ISSUE #16: Columnar Join and Sync Nodes Consequence
The columnar join function requires Thicket._sync_nodes_frame in order to match up the nodes in the dataframe after union operations have been performed in the graph. The function makes an assumption if a df node's frame matches a graph node's frame to set the df node to be the graph node, making their nid's equal. This results in some nodes being mapped to the same nid, if it's frame has already existed in the df.

image

I ran into this issue when traversing a graphs children and then trying to match those nodes in the dataframe. The problem appears for nodes like the one above because the same node frame appears for every kernel node that exists (obviously the kernel nodes have different frames, "apps_vol" vs "apps_hydro", etc).

Treece-Burgess:
This PR addresses adding the Pandas groupby functionality to the stats functions. This was done due to possible values being incorrectly ordered when stored within the aggregated statistic table. 

Fixes #16

### DIFF
--- a/thicket/stats/calc_boxplot_statistics.py
+++ b/thicket/stats/calc_boxplot_statistics.py
@@ -53,8 +53,9 @@ def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **
                 col + "_outliers" + q_list: [],
             }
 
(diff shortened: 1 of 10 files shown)


michaelmckinsey1:
depends on #43

PR #89, opened 2023-09-08
Improve Compatibility in concat_thickets

michaelmckinsey1:
# Description
Allow inner index to have any name, as long as it matches the index name in the metadata. Before this was limited to being named "profile" which is incompatible after using a function like `make_superthicket(profile_from_meta=x)`.


slabasan:
@MichaelMcKinsey1 I might have merged this out-of-order, when you get a chance can you resolve the conflicts?

slabasan (approved):

PR #100, opened 2023-10-13
Adds code to conf.py to fetch and use material from thicket-tutorial

ilumsden:
This PR adds code to `docs/conf.py` to fetch, run, and use the notebooks from thicket-tutorial when building the Thicket docs with Sphinx. `conf.py` now contains 3 variables to track content we want:
* `tutorial_target_branch`: a string which specifies the branch of [thicket-tutorial](https://github.com/LLNL/thicket-tutorial) to use
* `tutorial_notebooks`: a dictionary where keys are the paths to the desired notebooks relative to the root of [thicket-tutorial](https://github.com/LLNL/thicket-tutorial) and values are paths to where the executed notebooks should be dumped
* `extra_python_packages`: a list of extra PyPI packages to be appended to the end of the `requirements.txt` file from [thicket-tutorial](https://github.com/LLNL/thicket-tutorial)

The new code will then do the following:
* Clone [thicket-tutorial](https://github.com/LLNL/thicket-tutorial) into a temporary directory created with `tempfile` from the Python standard library
* Copy `requirements.txt` from the clone into Thicket's `docs` directory
* Append the contents of `extra_python_packages` to the copied `requirements.txt`
* Iterate through each notebook in `tutorial_notebooks`, run the notebook through [Papermill](https://papermill.readthedocs.io/en/latest/), and dump the executed notebook into its destination from `tutorial_notebooks`

This PR also includes minor changes to the Makefile in `docs` to delete executed notebooks and `requirements.txt` when a user runs `make clean`.

Finally, huge credit to @vanessalama09. Her work on a similar PR in thicket-tutorial (LLNL/thicket-tutorial#23) was integral in helping me come up with this idea.


ilumsden:
Unfortunately, we cannot use this because ReadTheDocs does not allow us to use a Docker container inside the Docker containers they use to build the documentation.

PR #106, opened 2023-10-30
debug github CI

slabasan:
Python 3.7 with ubuntu 20.04 was passing, then started failing tests in the last month. Using this PR to test some fixes.


slabasan:
Unclear how to support python 3.7 in GitHub CI, where newest pandas that can be installed is <1.4.0. Pandas 1.4.0 changes behavior for a shallow copy. See https://github.com/pandas-dev/pandas/issues/47703. The same behavior used to exist in pandas 1.2.5, but tests stopped passing.

PR #108, opened 2023-11-01
Fixes several bugs in reader_dispatch

ilumsden:
This PR fixes two bugs in `Thicket.reader_dispatch`.

First, it fixes type checking in the function. Currently, `reader_dispatch` uses `type(var) = expected_type`. This syntax may work in some cases. However, there is no guarantee it will work for all types, and the Python Software Foundation has reserved the right to completely invalidate this syntax in the future. This PR fixes this type checking by switching it to the more correct `isinstance(var, expected_type)` syntax.

Additionally, this PR fixes a bug in `Thicket.reader_dispatch` that would cause extra positional and keyword arguments to not be forwarded to the corresponding Hatchet function if the user passed in a list or tuple of filenames. This is particularly problematic for `Thicket.from_caliper` because it means the `query` argument is not forwarded to Hatchet, causing `Thicket.from_caliper` to fail in all cases.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -224,23 +224,32 @@ class Thicket(GraphFrame):
         """
         ens_list = []
         obj = args[0]  # First arg should be readable object
+        extra_args = []
+        if len(args) > 1:
+            extra_args = args[1:]
 
         # Parse the input object
         # if a list of files
-        if type(obj) == list:
+        if isinstance(obj, (list, tuple)):
             for file in obj:
-                ens_list.append(Thicket.thicketize_graphframe(func(file), file))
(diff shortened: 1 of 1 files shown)


michaelmckinsey1 (approved):

michaelmckinsey1:
In regards to type checking, I found this [reference](https://docs.quantifiedcode.com/python-anti-patterns/readability/do_not_compare_types_use_isinstance.html) that enforces the new solution implemented here. Most importantly, `isinstance` supports inheritance, whereas `type` does not. Using `type` to check types is an anti-pattern.

slabasan (approved):

PR #110, opened 2023-11-02
Andrew updated violin plots

AndyM1098:
Created two functions in order to plot violin plots.

The first function is designed to take in a singular thicket object, and plot user defined nodes and columns. The x-axis would be each node, and the y-axis would be the column metric specified by the user

The second function is designed to take in more than one thicket object, were we plot a node for each thicket, and the y-axis would be the data for each node. There are a few key things that are enforced. First, all thickets passed in needs to have the same columnar index level. Each thicket only gets one node to plot.

Both functions support custom percentile lines that would be user defined, as well as line styles and line colors. 

This PR needs to go after PR #103

### DIFF
--- a/thicket/stats/display_violinplot.py
+++ b/thicket/stats/display_violinplot.py
@@ -12,14 +12,26 @@ import matplotlib as mpl
 from .percentiles import percentiles
 from ..utils import verify_thicket_structures
 
+def _column_name_mapper(current_cols):
+    """
+        Internal function that returns string representation of current_cols
+
+        Parameters:
+            current_cols: 
+    """
+    if current_cols[0] in ["node", "name"]:
+        return current_cols[0]
+
+    return str(current_cols)
+
 def _add_percentile_lines_node(graph, thicket, nodes, columns, percentiles_vals, lines_styles=None, line_colors = None):
 
     violin_idx = -1
 
     #Default line styles and line colors
-    if lines_styles == None or len(lines_styles) == 0:
+    if lines_styles == None or len(lines_styles) != len(nodes):
         lines_styles = ["-"] * len(percentiles_vals)
-    if line_colors == None or len(line_colors) == 0:
+    if line_colors == None or len(line_colors) != len(nodes):
         line_colors = ["black"] * len(percentiles_vals)
 
     for node in nodes:
@@ -27,8 +39,10 @@ def _add_percentile_lines_node(graph, thicket, nodes, columns, percentiles_vals,
             violin_idx += 1
             for idx, percentile in enumerate(percentiles_vals):
                 stats_column = None
+                #Columnar joined thickets
                 if type(column) == type(tuple()):
                     stats_column = (str(column[0]), "{}_percentiles_{}".format(column[1], int(percentile * 100)))
+                #Non-columnar joined
                 else:
                     stats_column = "{}_percentiles_{}".format(column, int(percentile * 100))
                 #Call percentile(...) if the percentile value for the column has not been calculated already
@@ -36,6 +50,7 @@ def _add_percentile_lines_node(graph, thicket, nodes, columns, percentiles_vals,
                     or\
                     column in thicket.inc_metrics and stats_column not in thicket.statsframe.inc_metrics:
                         percentiles(thicket, [column], [percentile])
+
                 percentile_value = thicket.statsframe.dataframe[stats_column][node]
 
                 #Plot line
@@ -98,22 +113,10 @@ def display_violinplot(thicket, nodes=[], columns=[], percentiles = [], linestyl
         )
     # columnar joined thicket object
     else:
-
-        """
-            New code that allows for columns to be selected from different indexes of a columnar joined thicket
-        """
-        def column_name_mapper(current_cols):
-            if current_cols[0] in ["node", "name"]:
-                return current_cols[0]
-
-            return str(current_cols)
-
         cols = [str(c) for c in columns]
         df_subset = thicket.dataframe[[("name", ""), *columns]].reset_index()
-        df_subset.columns = df_subset.columns.to_flat_index().map(column_name_mapper)
+        df_subset.columns = df_subset.columns.to_flat_index().map(_column_name_mapper)
         df_subset["name"] = thicket.dataframe["name"].tolist()
-        # End new code
-        
         df = pd.melt(
             df_subset,
             id_vars=["node", "name"],
@@ -121,7 +124,6 @@ def display_violinplot(thicket, nodes=[], columns=[], percentiles = [], linestyl
             var_name="Performance counter",
             value_name=" ",
         )
-
         position = []
         for node in nodes:
             idx = df.index[df["node"] == node]
--- a/thicket/stats/display_violinplot.py
+++ b/thicket/stats/display_violinplot.py
@@ -35,7 +35,9 @@ def _add_percentile_lines(
     line_styles=None,
     line_colors=None,
 ):
-
+    # Error checking for types passed in by user:
+    # If percentiles is a list, then lines styles must be either a single value, or
+    # a list of values that match the length of the percentiles list. Same with line colors
     if isinstance(percentiles_vals, list):
         if isinstance(line_styles, list) is True:
             if len(percentiles_vals) != len(line_styles):
@@ -136,6 +138,7 @@ def _add_percentile_lines(
                         color=line_colors[idx],
                         linestyle=line_styles[idx],
                     )
+                    
                     # Get the violin structure
                     patch = mpl.patches.PathPatch(
                         graph.collections[violin_idx].get_paths()[0],
--- a/thicket/stats/display_violinplot.py
+++ b/thicket/stats/display_violinplot.py
@@ -29,6 +29,8 @@ def _add_percentile_lines_node(graph, thicket, nodes, columns, percentiles_vals,
                 stats_column = None
                 if type(column) == type(tuple()):
                     stats_column = (str(column[0]), "{}_percentiles_{}".format(column[1], int(percentile * 100)))
+                else:
+                    stats_column = "{}_percentiles_{}".format(column, int(percentile * 100))
                 #Call percentile(...) if the percentile value for the column has not been calculated already
                 if column in thicket.exc_metrics and stats_column not in thicket.statsframe.exc_metrics\
                     or\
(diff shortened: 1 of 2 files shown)


thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+    return graph
+
+
+def display_violinplot(
+    thicket,
+    nodes=[],
+    columns=[],
+    percentiles=[],
+    percentile_linestyles=[],
+    percentile_colors=[],
```
ilumsden: Change the accepted values of `percentiles`, `percentile_linestyles`, and `percentile_colors` as follows:

`percentiles` can be one of three values:
1. `None` = don't plot percentile lines
2. A single `float` = plot a single percentile line
3. A list of `float` = plot multiple percentile lines

`percentile_linestyle` and `percentile_colors` can be one of the following:
1. `None` = use default linestyles or colors
2. A single valid value for MPL's `linestyle` or `color` parameters = use the specified linestyle or color for every percentile line
3. A list of valid values for MPL's `linestyle` or `color` = use the specified linestyles of color for each percentile line. There needs to be a size check in this case to ensure the length of the list equals the length of the `percentiles` list

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                    column in thicket.exc_metrics
+                    and stats_column not in thicket.statsframe.exc_metrics
+                    or column in thicket.inc_metrics
+                    and stats_column not in thicket.statsframe.inc_metrics
+                ):
+                    percentiles(thicket, [column], [percentile])
+
+                percentile_value = thicket.statsframe.dataframe[stats_column][node]
+
+                # Plot line
```
AndyM1098: Add link to plotting line reference

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                    stats_column = "{}_percentiles_{}".format(
+                        column, int(percentile * 100)
+                    )
+                # Call percentile(...) if the percentile value for the column has not been calculated already
+                if (
+                    column in thicket.exc_metrics
+                    and stats_column not in thicket.statsframe.exc_metrics
+                    or column in thicket.inc_metrics
+                    and stats_column not in thicket.statsframe.inc_metrics
+                ):
```
ilumsden: Replace this with a check on the `dataframe` columns instead of `exc_metrics` and `inc_metrics`

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                        "{}_percentiles_{}".format(column[1], int(percentile * 100)),
+                    )
+                # Non-columnar joined
+                else:
+                    stats_column = "{}_percentiles_{}".format(
+                        column, int(percentile * 100)
+                    )
+                # Call percentile(...) if the percentile value for the column has not been calculated already
+                if (
+                    column in thicket.exc_metrics
```
AndyM1098: Change to check if column is in the actual statsframe, and not the list of metrics

thicket/stats/display_violinplot.py
ilumsden: Add a short comment that includes the link to StackOverflow where we found the code to do this

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+            percentile_colors,
+        )
+    # User did not specify percentiles, just return violinplot
+    else:
+        if len(columns) > 1:
+            return sns.violinplot(
+                data=filtered_df, x="node", y=" ", hue="Performance counter", **kwargs
+            )
+        else:
+            return sns.violinplot(data=filtered_df, x="node", y=" ", **kwargs)
```
ilumsden: Restructure all of this to minimize code duplication. For example, you could use something like the following psuedo code
```
graph = None
if len(columns) > 1:
    graph = <code to plot columnar join>
else:
    graph = <code to plot non-columnar join>
if len(percentiles) != 0:
    return <code to add percentile lines>
return graph
```

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+def display_violinplot_thicket(
+    thickets,
+    nodes=None,
+    columns=None,
+    x_order=None,
+    percentiles=[],
+    percentile_linestyles=[],
+    percentile_colors=[],
```
ilumsden, line 278: See the comment above about `percentiles`, `percentile_linestyles`, and `percentile_colors`

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                added_line.set_clip_path(patch)
+
+    return graph
+
+
+def display_violinplot_thicket(
+    thickets,
+    nodes=None,
+    columns=None,
+    x_order=None,
```
ilumsden, line 275: We still need to decide whether it makes more sense for `thickets`, `nodes`, and `columns` to be dictionaries or lists.

Additionally, we could consider letting `nodes` be a value that we use internally to get the actual Hatchet `Node` objects we need. For example, we could let `nodes` be a string representing the name of a node. Or, we could let it be a `Callable` that we invoke to get the `Node` object.

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+):
+
+    violin_idx = -1
+
+    # Default line styles and line colors
+    # If the user does not specify enough, default the missing values
+    if lines_styles is None:
+        lines_styles = ["--"] * len(percentiles_vals)
+    elif len(lines_styles) < len(percentiles_vals):
+        lines_styles += ["--"] * (len(percentiles_vals) - len(lines_styles))
```
ilumsden: Change this based on the comments regarding `percentiles`, `percentile_linestyles`, and `percentile_colors`

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+        for node in nodes:
+            idx = df.index[df["node"] == node]
+            for pos in idx:
+                position.append(pos)
+
+        # rename columns such that the x-axis label is "node" and not "name", tick marks
+        # will be node names
+        filtered_df = df.loc[position].rename(
+            columns={"node": "hatchet node", "name": "node"}
+        )
```
ilumsden, line 243: Move the code that builds the `position` list and the code that creates `filtered_df` outside of this if-else to avoid excessive code duplication

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+    # If the user does not specify enough, default the missing values
+    if lines_styles is None:
+        lines_styles = ["--"] * len(percentiles_vals)
+    elif len(lines_styles) < len(percentiles_vals):
+        lines_styles += ["--"] * (len(percentiles_vals) - len(lines_styles))
+
+    if line_colors is None:
+        line_colors = ["black"] * len(percentiles_vals)
+    elif len(line_colors) < len(percentiles_vals):
+        line_colors += ["black"] * (len(percentiles_vals) - len(line_colors))
```
ilumsden: See comment for this same section in the other percentile line plotting function

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                    color=line_colors[idx],
+                    linestyle=lines_styles[idx],
+                )
+                # Get the violin structure
+                patch = mpl.patches.PathPatch(
+                    graph.collections[violin_idx].get_paths()[0],
+                    transform=graph.transData,
+                )
+                # Clip line to the violin structure
+                added_line.set_clip_path(patch)
```
ilumsden: Could the interior of this large for loop be abstracted into another internal function? That would avoid the large amounts of code duplication in this function and `_add_percentile_lines_node`.

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+
+def display_violinplot_thicket(
+    thickets,
+    nodes=None,
+    columns=None,
+    x_order=None,
+    percentiles=[],
+    percentile_linestyles=[],
+    percentile_colors=[],
+    **kwargs,
```
ilumsden, line 279: @slabasan could it be useful to utilize type hints here? This is a feature since Python 3.5. If we were to use it, the function signature would look something like this:

```python3
import typing

def display_violinplot_thicket(
    thickets: dict[typing.Any, thicket.Thicket], # using 'Any' since we don't have a predefined type for groupby keys
    nodes: dict[typing.Any, hatchet.node.Node],
    columns: dict[typing.Any, list[str]],
    x_order=None, # x_order is the first optional argument
    ...):
    ...
```

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+            "Nodes must be a dictionary, specifying a node for each Thicket passed in"
+        )
+
+    if isinstance(columns, dict) is False:
+        raise ValueError(
+            "columns must be a dictionary, specifying columns for each Thicket passed in"
+        )
+
+    if all(isinstance(c, list) for c in columns.values()) is False:
+        raise ValueError("Columns dictionary must have list of columns as values")
```
ilumsden, line 325: A lot of these type checks can be removed if we decide to use type hinting

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,478 @@
...
+                value_name=" ",
+            )
+
+            idx_c = melted_df.index[melted_df["node"] == node]
+            position = [p for p in idx_c]
+            filtered_dfs[thicket_idx] = melted_df.loc[position].rename(
+                columns={"node": "hatchet node", "name": "node"}
+            )
+            thicket_name = [str(key)] * len(filtered_dfs[thicket_idx]["node"])
+            filtered_dfs[thicket_idx]["node"] = thicket_name
```
ilumsden, line 435: Could some of this code be abstracted into a separate internal function that can be used by both `display_violinplot` functions?

thicket/stats/display_violinplot.py
ilumsden: Actually, just replicate what you did in `display_violinplot_thicket`. That's what I'm suggesting here.

ilumsden (changes requested):
Looks pretty good to start. There are definitely still some things we need to work through as a group for `display_violinplot_thicket`

thicket/stats/display_violinplot.py
AndyM1098: Make sure type hinting throws an error on wrong types being passed in

AndyM1098:
Get rid of new lines below statements like if else, and for loops

slabasan:
All changes moved to #127

PR #118, opened 2023-12-08
Print Tree on Performance Data

michaelmckinsey1:
# Summary
- `tree()` now operates on the `Thicket.dataframe`
- Unit tests have been added for `tree()`. They are also the first unit tests to use the newest version of the rajaperf data `rajaperf-july-2023`.

# Motivation
The current tree function operates on the `statsframe.dataframe`, which is convenient since it matches the structure that the underlying renderer (Hatchet) expects, since it will only have 1 index level. While convenient from the developer side, we have found this is not intuitive for users as they expect the tree function to use the `dataframe` instead.

This feature simulates the single-level structure by taking a slice on the dataframe. Supporting printing a tree on any form of a `Thicket.dataframe` (more than 1 index level) would require an additional feature to the renderer to support visualizing the multi-index data structure.


thicket/external/console.py
```diff
@@ -135,6 +136,8 @@ def render(self, roots, dataframe, **kwargs):
         if self.color is True:
             result += self.render_legend()
 
+        result += f"Slice: {self.slice_tuple}"
```
michaelmckinsey1: We should somehow display what the slice is on the printout, but adding it to the legend would require modifications to Hatchet source code. We should probably add it to the hatchet code, although that would mean adding this `slice_tuple` object to the hatchet renderer code, which it does not apply to.

thicket/thicket.py
```diff
@@ -583,9 +585,52 @@ def tree(
...
+            raise KeyError(
+                f"Either dataframe cannot be represented as a single index or provided slice, '{indicies}' results in a multi-index. See self.dataframe.loc[(slice(None),)+{indicies},:]"
+            )
+        # For tree legend
+        idx_dict = {
+            slice_df.index.names[k]: indicies[k] for k in range(len(indicies))
+        }
+
+        # Prep DataFrame by filling None rows in the "name" column with the node's name.
+        slice_df["name"] = [n.frame["name"] for n in slice_df.index.get_level_values("node")]
```
michaelmckinsey1: @slabasan I just found that the filling method for missing rows, which puts "None" values in the name column for the missing rows does not work with tree. We can either revert this for just the name column or keep it as it is useful for functions like `intersection`. The above line avoids the issue for the dataframe that is passed to tree.

We should make a decision on how we want to handle this and that should be a separate PR but for now it is in here.

thicket/external/console.py
```diff
@@ -140,6 +141,81 @@ def render(self, roots, dataframe, **kwargs):
         else:
             return result.encode("utf-8")
 
+    def render_legend(self):
```
slabasan, line 146: This function pulled directly from hatchet code, right? If so, let's add a comment.

thicket/tests/conftest.py
```diff
@@ -230,3 +231,11 @@ def literal_thickets():
     thickets = [tk, tk2, tk3]
 
     return thickets
+
+
+@pytest.fixture
+def rajaperf_july_2023(data_dir, tmpdir):
```
slabasan: Do we still need to reference july2023 for the rajaperf data? Are these datafiles you are adding here already in thicket (thought we had merged in a PR with new datasets)?

slabasan (commented):
Small changes, couple quick questions.

thicket/external/console.py
michaelmckinsey1: Every function in `console.py` is sourced from hatchet `console.py` with various differences including `render_legend`. If the function is verbatim hatchet code we can just invoke it without pulling it over since `ThicketRenderer` is a child of `ConsoleRenderer`. I added a comment to emphasize this https://github.com/LLNL/thicket/pull/118/commits/25038e257341efac6c5a7ec950a895e5ed2a3b28

thicket/tests/conftest.py
michaelmckinsey1: The PR is now updated to work with the unit test data that we merged. The merged data happened after this PR so thats why it was in here.

slabasan (approved):

PR #125, opened 2023-12-30
update test data with newer rajaperf

slabasan:


michaelmckinsey1:
~Depends on #128. Currently rebased on top of it.~

michaelmckinsey1:
~Also depends on https://github.com/LLNL/hatchet/pull/117 on the hatchet side then unit tests should pass.~

PR #129, opened 2024-01-10
Rework _unify

michaelmckinsey1:
# Summary
Rework the `_unify` function to leverage the `old_to_new` object returned by `graph.union()` to more cleanly keep track of node updates during the unification process. Currently we update everything afterwards using `_sync_nodes_frame`, which requires some assumptions about the structure of the data. These implicit assumptions have resulted in bugs from our users. By leveraging the node map returned by the union function, we do not have to make any assumptions, which will reduce the amount of bugs we encounter.

# Changes
- Rewrite the `_unify` function
- Remove `_sync_nodes` and `_sync_nodes_frame`
- Add unify unit test
- Add literal reader for unit test

# Performance Impact
Where `n=len(thickets)`:
The current code unify is approx `O(n+n)` for computing unions `n` and sync nodes time (approx `n`). The new code adds a loop in each union step that makes the union block now `O(n^2)`. new-v2 and new-v3 are optimizations on the naive implementation of "new".

Small study
| files | develop | new | new-v2 | new-v3 (latest commit) |
|---|---|---|---|---|
| 10 | 0.16s | 0.22s | 0.2s | 0.15s |
50 | 0.8s | 2.2s | 1.2s | 0.77s
172 | 2.7s | 20.1s | 8.2s | 3.35s |
656 | 10.7s | 4m | 1m | 22s |
1438 | 24.3s | 21m | 6m | 1m21s |
1614 | 28s | 26m | 8m | 1m39s |
1808 | 31.3s | ? | 9m | 2m3s |

Current data line of best fit is linear (`Y = 0.07X + -1.73`)
New data line of best fit is nlog(n) (`Y = -123 + .5X*log_10(X)`). Empirically this is nlog(n) but if we had a worst case example at each size with n distinct graphs (improbable) I imagine it could actually fit n^2.


michaelmckinsey1:
Resolves #16

slabasan (approved):

PR #131, opened 2024-01-29
Add Progress Bar to Reader

michaelmckinsey1:
# Summary
Use `tqdm` to add a progress bar to the thicket reader since for larger amount of files (~1000+) in can take in the magnitude of minutes without any indication to the user (we don't want it to seem like it's hanging). The progress bar is in 2 stages for reading the files, and creating the thickets.

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,6 +46,7 @@ setup(
         "numpy",
         "pandas >= 1.1",
         "llnl-hatchet",
+        "tqdm",
     ],
     extras_require={
         "extrap": ["extrap", "matplotlib"],
--- a/thicket/ensemble.py
+++ b/thicket/ensemble.py
@@ -224,9 +224,9 @@ class Ensemble:
             combined_th.profile = [new_mappings[prf] for prf in combined_th.profile]
             profile_mapping_cp = combined_th.profile_mapping.copy()
             for k, v in profile_mapping_cp.items():
-                combined_th.profile_mapping[new_mappings[k]] = (
-                    combined_th.profile_mapping.pop(k)
-                )
+                combined_th.profile_mapping[
+                    new_mappings[k]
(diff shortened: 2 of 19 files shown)


michaelmckinsey1:
~Need to remove second stage and add flag to turn off tqdm by default~

slabasan:
@michaelmckinsey1 Two things: (1) Can you take a look at the conflicting file? (2) What do you think about setting `disable_tqdm=True` in our unit tests? For one, it'll show an example of using the new parameter, and also unit tests are small enough that we don't need to be showing progress. Not sure how this will render in our static jupyter notebook outputs in RTD, but maybe we disable it there as well.

slabasan (changes requested):
@michaelmckinsey1 Two things: (1) Can you take a look at the conflicting file? (2) What do you think about setting disable_tqdm=True in our unit tests? For one, it'll show an example of using the new parameter, and also unit tests are small enough that we don't need to be showing progress. Not sure how this will render in our static jupyter notebook outputs in RTD, but maybe we disable it there as well.

michaelmckinsey1:
> @michaelmckinsey1 Two things: (1) Can you take a look at the conflicting file? (2) What do you think about setting disable_tqdm=True in our unit tests? For one, it'll show an example of using the new parameter, and also unit tests are small enough that we don't need to be showing progress. Not sure how this will render in our static jupyter notebook outputs in RTD, but maybe we disable it there as well.

1. Fixed
2. ~Would we rather have it manually set in every test that uses a reader or `concat_thickets` or done like in https://github.com/LLNL/thicket/pull/131/commits/d92a8a17e064f11fc7b2340df8565cdcd0673205~ done manually 4fea8f3

slabasan (approved):

PR #136, opened 2024-02-26
Prevent test failure if Extra-P is not installed

ilumsden:
This simple PR adds a `pytest.skip` command that will skip over the Extra-P tests if Extra-P is not installed. This allows developers to run unit tests locally without needing to install Extra-P

### DIFF
--- a/thicket/tests/test_model_extrap.py
+++ b/thicket/tests/test_model_extrap.py
@@ -21,7 +21,9 @@ except ImportError:
     extrap_avail = False
 
 if not sys.version_info < (3, 8):
-    pytest.skip("requires python3.8 or greater to use extrap module", allow_module_level=True)
+    pytest.skip(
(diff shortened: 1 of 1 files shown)


slabasan (commented):
@ilumsden Question for my own understanding. We had initially put skips on the two extrap tests here if the python version was too old, why weren't these enough to prevent test failures?

ilumsden:
Those checks will only prevent test failures in cases where you absolutely cannot install Extra-P. So, for unit testing, that's more than enough.

However, for more day-to-day local testing, the existing Python version check would still require everyone else to install Extra-P wherever they're doing development to avoid test failures. In other words, if you have a new enough version of Python but you don't have Extra-P (which is common for local testing), the tests will fail because Python won't be able to find Extra-P.

slabasan (approved):

PR #141, opened 2024-03-06
Adds violin plot functionality along with testing

Yejashi:
Created two functions in order to plot violin plots.

The first function is designed to take in a singular thicket object, and plot user defined nodes and columns. The x-axis would be each node, and the y-axis would be the column metric specified by the user

The second function is designed to take in more than one thicket object, were we plot a node for each thicket, and the y-axis would be the data for each node. There are a few key things that are enforced. First, all thickets passed in needs to have the same columnar index level. Each thicket only gets one node to plot.

Both functions support custom percentile lines that would be user defined, as well as line styles and line colors.


thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,496 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+import pandas as pd
+import seaborn as sns
+import hatchet as ht
+import thicket as th
+import matplotlib as mpl
+from .percentiles import percentiles
+from ..utils import verify_thicket_structures
```
slabasan, line 13: I like following PEP8 for grouping imports:
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
```

So let's update this list to be:
```
import pandas as pd
import seaborn as sns
import matplotlib as mpl
import hatchet as ht

import thicket as th
from .percentiles import percentiles
from ..utils import verify_thicket_structures
```

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+def _column_name_mapper(current_cols):
+    """
+    Internal function that returns string representation of current_cols
+
+    Parameters:
+        current_cols:
```
slabasan: What type of input should `current_cols` be? Can you add this to the docstring?

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,496 @@
...
+    graphType,
+    thickets,
+    nodes,
+    columns,
+    x_order,
+    percentiles_vals,
+    line_styles=None,
+    line_colors=None,
+):
+
```
slabasan: You can remove the new line between the function definition and function body.

thicket/stats/display_violinplot.py
```diff
@@ -0,0 +1,496 @@
...
+            line styles will be "--". If the list does not specify styles for each of the percentiles, the missing
+            styles will default to "--"
+        percentile_colors (list): List of colors to apply for percentile lines. The default value is "black",
+            If the list does not specify colors for each percentile, the missing colors will default to "black"
+        **kwargs: Arguments that can be passed to seaborns plotting function
+
+    Returns:
+        (matplotlib Axes): Object for managing violinplot.
+    """
+
```
slabasan: New line here can be removed.

slabasan (changes requested):
Looks good overall @Yejashi! Just a couple small changes.

slabasan (approved):

PR #142, opened 2024-03-06
Improve Validation Checks

michaelmckinsey1:
# Summary
Currently if a unification operation results in a duplicate index in the perfdata multi-index, users will hit an error during the `_fill_perfdata` function. By moving the validation check before this function, which checks for duplicates, we can provide a clearer error message, which results in faster debugging for us and the user.

# Example
### For cases without metadata_key
```
DuplicateIndexError: Duplicate index {798655681} found in DataFrame index.
```
### For cases with metadata_key
```
DuplicateValueError: Different Thicket.metadata[metadata_key] may not contain duplicate values.
thickets[0].metadata['tuning'] and thickets[1].metadata['tuning'] have the same values: [{'block_128'}]
```
## Old
```
ValueError: cannot handle a non-unique multi-index!
```

### DIFF
--- a/thicket/ensemble.py
+++ b/thicket/ensemble.py
@@ -16,7 +16,6 @@ from .utils import (
     validate_dataframe,
     verify_sorted_profile,
     verify_thicket_structures,
-    DuplicateIndexError,
 )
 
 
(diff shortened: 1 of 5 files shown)


michaelmckinsey1:
~Depends on #125 unit test data~

slabasan (approved):

PR #148, opened 2024-04-04
Fix filter_stats

### LINKED ISSUE #147: filter_stats does not correctly update index in statsframe
`Thicket.filter_stats` will filter the Thicket object based on a predicate applied to the DataFrame in `Thicket.statsframe`. Currently, this function uses `hatchet.GraphFrame.squash` to propagate the filter to other components of the Thicket object (e.g., performance data and graph).

However, because Hatchet has no knowledge of the statsframe, `hatchet.GraphFrame.squash` will not properly update the index of `Thicket.statsframe.dataframe` with `Node` objects from the newly created graph.

As a result, the output of `Thicket.filter_stats` will look alright (i.e., the graph will contain the correct nodes and both the performance and stats dataframes will look correct), but functions that depend on the index of the statsframe (e.g., `Thicket.tree`) will fail with `IndexError`s.

michaelmckinsey1:
# Summary
This PR addresses the bug found in https://github.com/LLNL/thicket/issues/147. The fix leverages the `old_to_new` mapping created in `squash` to synchronize the node objects in the `statsframe.dataframe` with the `graph` and `dataframe`. This introduces a new capability in `Thicket.squash` to update the statsframe instead of creating a new one every time (`new_statsframe=False/True`). Additionally, a utils function is added and called in `filter_stats` to check for the invalid state observed in the bug. This PR also migrates hatchet squash code to Thicket squash to expose the `old_to_new` object and is a part of a decoupling effort for `Thicket` dependencies on `Hatchet`.

Fixes #147 

Example output from the utility:
image

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -25,7 +25,7 @@ from .utils import (
     verify_thicket_structures,
     check_duplicate_metadata_key,
     validate_profile,
-    validate_nodes
+    validate_nodes,
 )
 from .external.console import ThicketRenderer
 
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -8,12 +8,13 @@ import os
 import sys
 import json
 import warnings
-from collections import OrderedDict
+from collections import defaultdict, OrderedDict
 from hashlib import md5
 
 import pandas as pd
 import numpy as np
 from hatchet import GraphFrame
+from hatchet.graph import Graph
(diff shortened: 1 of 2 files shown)


michaelmckinsey1:
~depends on https://github.com/LLNL/hatchet/pull/121~

slabasan:
@michaelmckinsey1 Can you fix conflicts in `thicket.py`?

slabasan (approved):

PR #156, opened 2024-05-01
Refinements to display_violinplot_thicket Function

Yejashi:
This PR improves the display_violinplot_thicket function in two ways:

- Changed default xlabel to 'thicket' for clarity.

- Restricted function to accept only the same node across all provided thickets.

### DIFF
--- a/thicket/stats/display_violinplot.py
+++ b/thicket/stats/display_violinplot.py
@@ -381,6 +381,12 @@ def display_violinplot_thicket(
             "Value(s) passed to 'nodes' argument must be of type hatchet.node.Node."
         )
 
+    # Verify that all nodes passed are of the same type and name
+    if len(set(nodes.values())) != 1:
+        raise ValueError(
+            "Value(s) passed to 'nodes' argument must be of same type and name."
+        )
+
     # the keys must match the thickets dictionary keys
     if set(thickets.keys()) != set(nodes.keys()):
         raise ValueError(
@@ -476,9 +482,15 @@ def display_violinplot_thicket(
             filtered_dfs[thicket_idx]["node"] = thicket_name
 
     master_df = pd.concat(filtered_dfs, ignore_index=True)
+    master_df.rename(columns={"node": "thicket"}, inplace=True)
 
     graph = sns.violinplot(
-        data=master_df, x="node", y=" ", hue="Performance counter", inner=None, **kwargs
+        data=master_df,
+        x="thicket",
+        y=" ",
+        hue="Performance counter",
+        inner=None,
+        **kwargs,
     )
 
     # No percentiles to plot!
--- a/thicket/stats/display_violinplot.py
+++ b/thicket/stats/display_violinplot.py
@@ -382,7 +382,7 @@ def display_violinplot_thicket(
         )
 
     # Verify that all nodes passed are of the same type and name
-    if len(set(nodes.values())) != 1:
+    if len(set([n.frame for n in nodes.values()])) != 1:
         raise ValueError(
             "Value(s) passed to 'nodes' argument must be of same type and name."
         )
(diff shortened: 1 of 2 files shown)


thicket/stats/display_violinplot.py
```diff
@@ -381,6 +381,12 @@ def display_violinplot_thicket(
             "Value(s) passed to 'nodes' argument must be of type hatchet.node.Node."
         )
 
+    # Verify that all nodes passed are of the same type and name
+    if len(set(nodes.values())) != 1:
+        raise ValueError(
+            "Value(s) passed to 'nodes' argument must be of same type and name."
+        )
+
```
ilumsden, line 389: This is likely to cause a lot of errors. To determine if objects are present, sets compute hashes. If you look at [Hatchet's Node class](https://github.com/LLNL/hatchet/blob/6a6d7027056df96bd1c919ab34a9acce81f3b9a1/hatchet/node.py#L190), the hash of a `Node` object (i.e., the type of the objects returned from `nodes.values()`) is just the `_hatchet_nid` field. There is currently no guarantee that `Node` objects representing the same region in two Thickets will have the same `_hatchet_nid` value. For example, a `Node` for e.g., `MPI_Send` could have a `_hatchet_nid` of 5 in one Thicket and 8 in the other.

To fix this, you need to put the `Frame` objects in the set. You can do that with:
```python
if len(set([n.frame for n in nodes.values()])):
```

thicket/tests/test_display.py
```diff
@@ -336,8 +336,8 @@ def test_display_violinplot_thicket(thicket_axis_columns):
         "th_2": ["Min time/rank"],
         "th_3": ["Min time/rank"],
     }
-    node = pd.unique(combined_th.dataframe.reset_index()["node"][0:1]).tolist()[0]
-    nodes = {"th_1": node, "th_2": node, "th_3": node}
+    node = pd.unique(combined_th.dataframe.reset_index()["node"][:]).tolist()
+    nodes = {"th_1": node[0], "th_2": node[0], "th_3": node[0]}
```
ilumsden: This should not work. The nodes passed to `display_violinplot_thicket` should come from the actual Thickets passed to the function. Here, you are getting the nodes from one Thicket (`combined_th`), but you are passing different Thickets (from `thicket_list`) to `display_violinplot_thicket`.

To fix this, you need to get your nodes from `thicket_list[0]`, `thicket_list[1]`, and `thicket_list[2]`.

thicket/tests/test_display.py
```diff
@@ -364,6 +364,15 @@ def test_display_violinplot_thicket(thicket_axis_columns):
             thickets=thickets, nodes=["test"], columns=columns
         )
 
+    with pytest.raises(
+        ValueError,
+        match=r"Value\(s\) passed to 'nodes' argument must be of same type and name.",
+    ):
+        invalid_nodes = {"th_1": node[1], "th_2": node[0], "th_3": node[0]}
```
ilumsden: See comment about which nodes you need to capture

thicket/tests/test_display.py
```diff
@@ -423,7 +432,7 @@ def test_display_violinplot_thicket(thicket_axis_columns):
         ValueError,
         match="Keys in 'nodes' argument dictionary do not match keys in 'thickets' argument dictionary.",
     ):
-        false_nodes = {"th_x": node, "th_2": node, "th_3": node}
+        false_nodes = {"th_x": node[0], "th_2": node[0], "th_3": node[0]}
```
ilumsden: See comment about which nodes you need to capture

ilumsden (changes requested):
Most of this looks good, but there are two main issues that need to be addressed before I can approve.

ilumsden:
@Yejashi please rebase this PR on #160 or rebase on `develop` after #160 is merged. There is a bug in the re-exports of `display_violinplot.py` that is addressed by #160. If you do not merge the changes from this PR, you will reintroduce the bug.

ilumsden (approved):
This looks good. @pearce8 just don't merge until my comment regarding #160 is addressed

pearce8 (approved):

PR #157, opened 2024-05-02
Implementation of query_stats function

Yejashi:
This PR implements the query_stats function, introducing query language functionality to the statsframe. This enhancement allows end users to perform queries based on data within the statsframe.

### Summary of Changes

To achieve this, extensive modifications were necessary. Specifically, all existing stats functions now return a list of output column names. This adjustment was essential to support the caching functionality required for the query_stats function.

#### Further Explanation

When the query_stats function is called with the squash flag enabled, nodes are sometimes merged into one. This creates two primary issues:

1.  A mismatch between the expected number of nodes in the performance data and the statsframe.
2.  Invalid data within the statsframe, as some nodes have been merged and their values aggregated.

To address these issues, the statsframe needs to be recalculated after a squash operation. To facilitate this recalculation, I added a caching mechanism. This mechanism records the most recent stats operations and stores them in `statsframe_ops_cache`. These cached operations are then reapplied on a new statsframe.

By implementing this caching functionality, we ensure that the statsframe remains consistent and accurate after squash operations, enabling reliable query results.

### Important Notice

For the methods in this PR to work in the future, we must ensure a standard across all stats function such that they must return a list of output column names. This is an integral aspect of the caching mechanism.


thicket/thicket.py
```diff
@@ -1173,6 +1173,67 @@ def query(
             return filtered_th.squash(update_inc_cols=update_inc_cols)
         return filtered_th
 
+    def query_stats(
+        self, query_obj, squash=True, update_inc_cols=True, multi_index_mode="off"
```
ilumsden: The `multi_index_mode` argument should not be needed for `query_stats`. As of now, the statsframe should only have a one-level row index. `multi_index_mode` is only needed when the DataFrame used in the querying has a multi-level row index.

thicket/thicket.py
```diff
@@ -1173,6 +1173,67 @@ def query(
...
+            query, self.statsframe.graph, self.statsframe.dataframe
+        )
+        filtered_sf_df = sframe_copy.loc[sframe_copy["node"].isin(query_matches)]
+
+        if filtered_sf_df.shape[0] == 0:
+            raise EmptyQuery("The provided query would have produced an empty Thicket.")
+
+        index_names = self.dataframe.index.names
+        dframe_copy = self.dataframe.copy()
+        index_names = self.dataframe.index.names
```
ilumsden: Why is this line here twice?

thicket/thicket.py
```diff
@@ -1173,6 +1173,67 @@ def query(
...
+        dframe_copy.reset_index(inplace=True)
+
+        filtered_df = dframe_copy[dframe_copy["node"].isin(query_matches)]
+        filtered_df.set_index(index_names, inplace=True)
+
+        filtered_th = self.deepcopy()
+        filtered_th.dataframe = filtered_df
+
+        if squash:
+            return filtered_th.squash(update_inc_cols=update_inc_cols)
```
ilumsden: At this point, the new `Thicket` object will have:
* A correctly-filtered performance DataFrame
* A correctly-filtered Graph

However, you will have either an empty statsframe or the old statsframe (depending on whether `squash` clears the statsframe).

You need to add code to update the statsframe.

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,177 @@
...
+    filt_th_df_nodes = filt_th.dataframe.index.get_level_values(node_name).to_list()
+    filt_th_df_profiles = filt_th.dataframe.index.get_level_values(profile_name)
+
+    assert len(filt_nodes) == len(match)
+    assert all([n.frame in match_frames for n in filt_nodes])
+    assert all([n.frame["name"] in match_names for n in filt_nodes])
+    assert all([n in filt_th_df_nodes for n in filt_nodes])
+    assert sorted(filt_th_df_profiles.unique().to_list()) == sorted(
+        th_df_profiles.unique().to_list()
+    )
```
ilumsden, line 53: You are not checking anything regarding the statsframe here. You need to add code to make sure the statsframe is valid too.

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,177 @@
...
+        set(
+            [
+                root,  # RAJAPerf
+                root.children[3],  # RAJAPerf.Lcals
+                root.children[4],  # RAJAPerf.Polybench
+            ]
+        )
+    )
+
+    new_th = th_cj.query_stats(query, multi_index_mode="all")
```
ilumsden: You should not have `multi_index_mode` for `query_stats`

ilumsden (changes requested):
There's one major issue right now with `query_stats` and the corresponding tests:

They don't update or check the statsframe.

Since you are querying the statsframe, I would expect the output Thicket to contain the following:
1. A statsframe containing only nodes captured by the query 
2. A graph containing only nodes captured by the query
3. A performance dataframe containing only nodes captured by the query
4. An unmodified metadata table

Your code right now produces `Thicket` objects with elements 2, 3, and 4, but it doesn't add the filtered statsframe (i.e., element 1) back to the new `Thicket`.

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,192 @@
...
+    assert all([n.frame in match_frames for n in filt_nodes])
+    assert all([n.frame["name"] in match_names for n in filt_nodes])
+    assert all([n in filt_th_df_nodes for n in filt_nodes])
+    assert sorted(filt_th_df_profiles.unique().to_list()) == sorted(
+        th_df_profiles.unique().to_list()
+    )
+
+    assert len(sframe_nodes) == len(match)
+    assert all([n.frame in match_frames for n in sframe_nodes])
+    assert all([n.frame["name"] in match_names for n in sframe_nodes])
```
ilumsden, line 57: The check on line 50 (see below) should also be duplicated for statsframe:

```python
assert all([n in filt_th_df_nodes for n in filt_nodes])
```

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,192 @@
...
+                root.children[3],  # RAJAPerf.Lcals
+                root.children[4],  # RAJAPerf.Polybench
+            ]
+        )
+    )
+
+    new_th = th_cj.query_stats(query)
+
+    th.stats.mean(new_th, columns=[(0, "Min time/rank")])
+    th.stats.mean(new_th, columns=[(1, "Min time/rank")])
```
ilumsden: Why are you calling `th.stats.mean` again? The statsframe should not be cleared, so, unless I'm missing something, there should be no need to call `mean` again

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,192 @@
...
+    queried_nodes = list(new_th.graph.traverse())
+
+    # Get statsframe nodes
+    sframe_nodes = new_th.statsframe.dataframe.reset_index()["node"]
+
+    match_frames = list(sorted([n.frame for n in match]))
+    queried_frames = list(sorted([n.frame for n in queried_nodes]))
+
+    th.stats.mean(new_th, columns=[(0, "Min time/rank")])
+    th.stats.mean(new_th, columns=[(1, "Min time/rank")])
```
ilumsden: Again, why are you calling `th.stats.mean` here?

thicket/thicket.py
```diff
@@ -1173,6 +1173,71 @@ def query(
...
+
+        filtered_df = dframe_copy[dframe_copy["node"].isin(query_matches)]
+        filtered_df.set_index(df_index_names, inplace=True)
+
+        filtered_th = self.deepcopy()
+        filtered_th.dataframe = filtered_df
+
+        sframe_copy = sframe_copy[
+            sframe_copy["node"].apply(lambda x: x in query_matches)
+        ]
```
ilumsden: You've already done this in a more pandas-native, performant why on line 1213 (i.e., where you define `filtered_sf_df`). Remove this code and just use `filtered_sf_df`

thicket/thicket.py
```diff
@@ -1173,6 +1173,71 @@ def query(
...
+        sframe_copy.set_index(sf_index_names, inplace=True)
+        filtered_th.statsframe.dataframe = sframe_copy
```
ilumsden: If you're going to start with a deepcopy of the input `Thicket` object, you need to also update `filtered_th.statsframe.graph` so that it is the same as `filtered_th.graph`.

ilumsden (changes requested):
Very good, but there's still a few things that need to be addressed

thicket/stats/calc_boxplot_statistics.py
```diff
@@ -108,6 +117,10 @@ def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **
                 }
             }
 
+            output_column_names.append(
+                str((boxplot_dict.keys(), next(iter(boxplot_dict.values())).keys()))
+            )
```
ilumsden: This is really over-complicated for getting the metric names for multi-level columns. You need to loop over the keys of the inner dictionary and add `str((idx, key))` for each iteration. So:
```python
for k in boxplot_dict[idx]:
    output_column_names.append(str((idx, k)))
```

thicket/stats/maximum.py
```diff
@@ -43,6 +51,7 @@ def maximum(thicket, columns=None):
     else:
         df = thicket.dataframe[columns].reset_index(level=1).groupby("node").agg(max)
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_max")))
```
ilumsden: This applies to more than just this file, but you don't need to pass the tuple through `str()`. A tuple of strings is hashable, so it can be used as a key in the stats cache just as easily as its "stringified" version.

thicket/stats/percentiles.py
```diff
@@ -63,8 +70,10 @@ def percentiles(thicket, columns=None, percentiles=[0.25, 0.50, 0.75]):
             for node in pd.unique(df.reset_index()["node"].tolist()):
                 calculated_percentiles.append(list(df.loc[node][column]))
 
+            column_name_agg = ""
             for index, percentile in enumerate(percentiles):
                 column_to_append = column + "_percentiles_" + str(int(percentile * 100))
+                column_name_agg = column_name_agg + column_to_append
```
ilumsden: Do not do this. These are all separate columns, so there is no reason to concatenate their names together for the cache. Just do `output_column_names.append(column_to_append)`

thicket/stats/percentiles.py
```diff
@@ -99,6 +111,7 @@ def percentiles(thicket, columns=None, percentiles=[0.25, 0.50, 0.75]):
                     idx_level,
                     "{}_percentiles_{}".format(column, str(int(percentile * 100))),
                 )
+                column_name_agg = column_name_agg + str(column_to_append)
```
ilumsden: See previous comments about `str()` and concatenation

thicket/stats/preference.py
```diff
@@ -71,6 +78,10 @@ def preference(thicket, columns, comparison_func, test="ttest", *args, **kwargs)
         aggregated_cols = columns[0] + " vs " + columns[1]
         thicket.statsframe.dataframe[aggregated_cols + "_std_preferred"] = pref_std
         thicket.statsframe.dataframe[aggregated_cols + "_mean_preferred"] = pref_mean
+
+        output_column_names.append(
+            aggregated_cols + "_std_preferred" + aggregated_cols + "_mean_preferred"
```
ilumsden: Again, do not just concatenate these stats metric names. This produces two stats columns, so append two column names

thicket/stats/preference.py
```diff
@@ -96,16 +107,17 @@ def preference(thicket, columns, comparison_func, test="ttest", *args, **kwargs)
...
         thicket.statsframe.dataframe[
-            (
-                "Preference",
-                aggregated_cols + "_mean_preferred",
-            )
+            (col_name[0], col_name[1] + "_mean_preferred")
         ] = pref_mean
+
+        output_column_names.append(
+            str(col_name) + "_std_preferred" + str(col_name) + "_mean_preferred"
```
ilumsden: Again, no concatenation

thicket/stats/stats_utils.py
```diff
@@ -0,0 +1,13 @@
+def cache_decorator(func):
```
ilumsden: Please rename this. This is a really generic name. How about `cache_stats_op`?

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,188 @@
...
+        assert isinstance(filt_th.statsframe.dataframe.columns, pd.MultiIndex)
+
+    # Get filtered nodes and profiles
+    filt_th_df_nodes = filt_th.dataframe.index.get_level_values(node_name).to_list()
+    filt_th_df_profiles = filt_th.dataframe.index.get_level_values(profile_name)
+
+    assert len(filt_nodes) == len(match)
+    assert all([n.frame in match_frames for n in filt_nodes])
+    assert all([n.frame["name"] in match_names for n in filt_nodes])
+    assert all([n in filt_th_df_nodes for n in filt_nodes])
```
ilumsden: The original `assert` just checks if the nodes in the graph form a subset of the nodes in the performance dataframe. The performance dataframe could contain more nodes than the graph and still pass this test.

Instead, change this to:
```python
assert sorted(filt_nodes) == sorted(filt_th_df_nodes)
```

`Node` objects are comparable (comparison is w.r.t their `_hatchet_nid` fields). As a result, you can sort both lists and then use `==` to check that the set of nodes in the Graph and performance dataframe are truly equal.

Note that you can only do this type of comparison when all the nodes come from the same `Thicket` object. There is no guarantee that this will work across `Thicket` objects (e.g., the comparisons involving `match`, `match_frames`, and `match_names` above).

ilumsden: Actually, you may need to make sure that the contents of both are unique first. This won't work if there are duplicates in either list.

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,188 @@
...
+    assert sorted(filt_th_df_profiles.unique().to_list()) == sorted(
+        th_df_profiles.unique().to_list()
+    )
+
+    assert len(sframe_nodes) == len(match)
+    assert all([n.frame in match_frames for n in sframe_nodes])
+    assert all([n.frame["name"] in match_names for n in sframe_nodes])
+    # Only needed if new_statsframe=False
+    assert all([n in pd.unique(filt_th_df_nodes) for n in sframe_nodes])
```
ilumsden, line 58: See the comment above regarding using `==` instead of `all()`. Also, this line provides an example for how to make the lists unique first.

thicket/tests/test_stats.py
```diff
@@ -1096,3 +1096,33 @@ def test_score_hellinger(thicket_axis_columns):
...
+
+
+def test_cache_decorator(rajaperf_seq_O3_1M_cali):
+    th_1 = th.Thicket.from_caliperreader(rajaperf_seq_O3_1M_cali, disable_tqdm=False)
+
+    th.stats.mean(th_1, columns=["Min time/rank"])
+
+    assert len(th_1.statsframe_ops_cache) == 1
+    assert (
+        len(th_1.statsframe_ops_cache[list(th_1.statsframe_ops_cache.keys())[0]]) == 1
```
ilumsden, line 1127: There's no need for two tests here. Just copy this line into `test_reapply_statsframe_operations`.

thicket/thicket.py
```diff
@@ -396,6 +403,7 @@ def _index(thickets, from_statsframes=False, disable_tqdm=disable_tqdm):
                 metadata=thicket_parts[4],
                 profile=thicket_parts[5],
                 profile_mapping=thicket_parts[6],
+                statsframe_ops_cache={},
```
ilumsden, line 407: I'd recommend adding comments here and on line 418 (I'll add a comment there) saying why we reset the cache. Just a useful thing to make sure we don't forget in the future.

thicket/thicket.py
```diff
@@ -407,6 +415,7 @@ def _columns(
                 metadata_key=metadata_key,
                 disable_tqdm=disable_tqdm,
             )
+            combined_thicket.statsframe_ops_cache = {}
```
ilumsden, line 420: Add a comment here too

thicket/thicket.py
```diff
@@ -1173,6 +1185,97 @@ def query(
...
+        filtered_df = dframe_copy[dframe_copy["node"].isin(query_matches)]
+        filtered_df.set_index(df_index_names, inplace=True)
+
+        filtered_th = self.deepcopy()
+
+        filtered_sf_df.set_index(sf_index_names, inplace=True)
+        filtered_th.statsframe.dataframe = filtered_sf_df
+
+        if squash:
+            filtered_th.dataframe = filtered_df
```
ilumsden: This needs to happen even if we don't squash. If we change the stats data, we should change the perf data.

thicket/thicket.py
```diff
@@ -1173,6 +1185,97 @@ def query(
...
+            raise EmptyQuery("The provided query would have produced an empty Thicket.")
+
+        df_index_names = self.dataframe.index.names
+        dframe_copy = self.dataframe.copy()
+        dframe_copy.reset_index(inplace=True)
+
+        filtered_df = dframe_copy[dframe_copy["node"].isin(query_matches)]
+        filtered_df.set_index(df_index_names, inplace=True)
+
+        filtered_th = self.deepcopy()
```
ilumsden: Given how much we're changing, do we really need a `deepcopy` here? It might be better to just create a brand new `Thicket` object with the filtered statsframe and dataframe, the same graph, and other bookkeeping members copied.

ilumsden: @pearce8 I'd appreciate your feedback here too

thicket/thicket.py
```diff
@@ -1173,6 +1185,97 @@ def query(
...
+            filtered_th.dataframe = filtered_df
+
+            final_thicket = filtered_th.squash(
+                update_inc_cols=update_inc_cols, new_statsframe=True
+            )
+            final_thicket.statsframe.graph = final_thicket.graph
+
+            final_thicket.reapply_stats_operations()
+
+            return final_thicket
```
ilumsden: Once `filtered_th.dataframe = filtered_df` is moved outside of this conditional, we can just do all our operations on `filtered_th`. So, this return statement will become redundant.

thicket/thicket.py
```diff
@@ -1173,6 +1185,97 @@ def query(
...
+    def reapply_stats_operations(self):
+        """Reapply most recent stats operations."""
+
+        for stats_func, arg_dict in self.statsframe_ops_cache.items():
+            # This makes sure the data is comparable
+            sortkey = lambda x: (  # noqa: E731
+                x[0],
+                tuple(sorted(x[1].items(), key=lambda y: y[0])),
+            )
+            for k, g in groupby(
```
ilumsden: Given that we have a `Thicket.groupby` function, it would probably be better to change the import for groupby to just `import itertools` and then change this call to `for k, g in itertools.groupby(...):`

thicket/thicket.py
```diff
@@ -1173,6 +1185,97 @@ def query(
...
+                list(sorted(arg_dict.values(), key=sortkey)), key=sortkey
+            ):
+                arg = list(g)[0]
+                if arg[0] is None:
+                    stats_func(self, **arg[1])
+                    continue
+                elif arg[1] is None:
```
ilumsden: Will these ever be `None`? If not, these conditionals aren't needed. Otherwise, these conditions should probably just convert `arg[0]` to `()` and convert `arg[1]` to `{}`.

ilumsden (changes requested):
This is very close to being good to go. I've got a few more changes, but I suspect this PR will probably be good after that.

thicket/thicket.py
Yejashi: No that i think about it, there might not be a reason to reset the cache when a concat with axis set to index is performed. All that really changes is the nodes and not the column.

thicket/thicket.py
Yejashi: @ilumsden, suppose i created a new thicket in place of deepcopy, i still need to run squash on it so doesn't that mean there will be three thickets regardless?

thicket/thicket.py
michaelmckinsey1, line 407:
```suggestion
```
Why is this line necessary when it will be created in the constructor?

thicket/thicket.py
michaelmckinsey1: I don't see the difference of making a new Thicket as you would need to give the constructor copies of the graph and dataframe anyway.

ilumsden: The graph maybe, but you don't need to give it copies of the dataframes. Passing to a constructor will be pass by reference.

I told Befikir this at lunch, but the reason to avoid `deepcopy` is because you are creating copies of the performance dataframe and (potentially) stats dataframe only to immediately replace them. There's no reason to do that.

thicket/thicket.py
ilumsden: I'd argue this is fine. Like you said, it's not necessary, but it does convey the intent of the call more clearly than just leaving it out.

thicket/stats/calc_boxplot_statistics.py
```diff
@@ -51,6 +58,8 @@ def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **
                 col + "_outliers" + q_list: [],
             }
 
+            output_column_names.append(str(boxplot_dict.keys()))
```
ilumsden: You should loop over `boxplot_dict.keys()`. Right now, this code is creating an iterator of strings (although you can think of it like a list) and stringifying it. Instead, you want each key to be added to `output_column_names`

thicket/stats/calc_boxplot_statistics.py
```diff
@@ -108,6 +117,9 @@ def calc_boxplot_statistics(thicket, columns=[], quartiles=[0.25, 0.5, 0.75], **
                 }
             }
 
+            for k in boxplot_dict[idx]:
+                output_column_names.append(str((idx, k)))
```
ilumsden: Why are you stringifying the tuples for multi-indexed columns? Since the contents of the tuples are strings, you can use the `tuple` in the cache without stringification. Also, by stringifying, you render the return from this function unusable by users.

thicket/stats/check_normality.py
```diff
@@ -72,6 +80,7 @@ def check_normality(thicket, columns=None):
             .agg(stats.shapiro)
         )
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_normality")))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/correlation_nodewise.py
```diff
@@ -100,17 +109,22 @@ def correlation_nodewise(thicket, column1=None, column2=None, correlation="pears
...
-                    "Union statistics",
-                    column1[1] + "_vs_" + column2[1] + " " + correlation,
-                )
-            ] = correlated
+            column_name = (
+                "Union statistics",
+                column1[1] + "_vs_" + column2[1] + " " + correlation,
+            )
+            thicket.statsframe.dataframe[column_name] = correlated
+            output_column_names.append(str(column_name))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/correlation_nodewise.py
```diff
@@ -100,17 +109,22 @@ def correlation_nodewise(thicket, column1=None, column2=None, correlation="pears
...
             column_idx = column1[0]
-            thicket.statsframe.dataframe[
-                (column_idx, column1[1] + "_vs_" + column2[1] + " " + correlation)
-            ] = correlated
+            column_name = (
+                column_idx,
+                column1[1] + "_vs_" + column2[1] + " " + correlation,
+            )
+            thicket.statsframe.dataframe[column_name] = correlated
+            output_column_names.append(str(column_name))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/calc_boxplot_statistics.py
ilumsden: Actually, I see that you have removed the stringification in some places. Given that, just treat this comment (and the other ones that refer to this) as indicators of places where you missed removing it

thicket/stats/mean.py
```diff
@@ -44,6 +52,7 @@ def mean(thicket, columns=None):
             thicket.dataframe[columns].reset_index(level=1).groupby("node").agg(np.mean)
         )
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_mean")))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/median.py
```diff
@@ -48,6 +56,7 @@ def median(thicket, columns=None):
             .agg(np.median)
         )
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_median")))
```
ilumsden, line 59: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/minimum.py
```diff
@@ -42,6 +50,7 @@ def minimum(thicket, columns=None):
     else:
         df = thicket.dataframe[columns].reset_index(level=1).groupby("node").agg(min)
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_min")))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/preference.py
```diff
@@ -96,16 +106,16 @@ def preference(thicket, columns, comparison_func, test="ttest", *args, **kwargs)
...
         thicket.statsframe.dataframe[
-            (
-                "Preference",
-                aggregated_cols + "_mean_preferred",
-            )
+            (col_name[0], col_name[1] + "_mean_preferred")
         ] = pref_mean
+
+        output_column_names.append(str(col_name) + "_std_preferred")
+        output_column_names.append(str(col_name) + "_mean_preferred")
```
ilumsden: This is similar to the comment about stringification in `calc_boxplot_statistics.py`

When you use `str(col_name)`, you are stringifying the tuple and then adding `"_std_preferred"` or `"_mean_preferred"` to it. To fix this, just copy the indexing into `thicket.statsframe.dataframe` above. So, for example:
```python
str(col_name) + "_std_preferred"
```
would become:
```python
(col_name[0], col_name[1] + "_std_preferred")
```

thicket/stats/scoring.py
```diff
@@ -150,9 +152,12 @@ def score(thicket, columns, output_column_name, scoring_function):
     thicket.statsframe.dataframe["Scoring", stats_frame_column_name] = resulting_scores
     thicket.statsframe.dataframe = thicket.statsframe.dataframe.sort_index(axis=1)
 
-    return
+    output_column_names.append(str(("Scoring", stats_frame_column_name)))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/std.py
```diff
@@ -44,6 +52,7 @@ def std(thicket, columns=None):
     else:
         df = thicket.dataframe[columns].reset_index(level=1).groupby("node").agg(np.std)
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_std")))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/stats/variance.py
```diff
@@ -45,6 +53,7 @@ def variance(thicket, columns=None):
     else:
         df = thicket.dataframe[columns].reset_index(level=1).groupby("node").agg(np.var)
         for idx, column in columns:
+            output_column_names.append(str((idx, column + "_var")))
```
ilumsden: See comment about stringification in `calc_boxplot_statistics.py`

thicket/tests/test_query_stats.py
```diff
@@ -0,0 +1,188 @@
...
+    assert sorted(pd.unique(filt_th_df_nodes)) == sorted(pd.unique(filt_nodes))
+    assert sorted(filt_th_df_profiles.unique().to_list()) == sorted(
+        th_df_profiles.unique().to_list()
+    )
+
+    assert len(sframe_nodes) == len(match)
+    assert all([n.frame in match_frames for n in sframe_nodes])
+    assert all([n.frame["name"] in match_names for n in sframe_nodes])
+    assert all([n in pd.unique(filt_th_df_nodes) for n in sframe_nodes])
+    assert sorted(pd.unique(filt_th_df_nodes)) == sorted(pd.unique(sframe_nodes))
```
ilumsden, line 59: Since this is passing unit tests, there's no need to change this. But, for future reference, it's usually a good idea to make sure you aren't calling pandas functions on non-pandas types (e.g., `pd.unique(filt_th_df_nodes)`) or vice versa (e.g., `sorted(pd.unique(...))`)

thicket/thicket.py
```diff
@@ -1173,6 +1187,96 @@ def query(
...
+        filtered_sf_df.set_index(sf_index_names, inplace=True)
+
+        filtered_th = Thicket(
+            graph=self.graph,
+            dataframe=filtered_df,
+            exc_metrics=self.exc_metrics.copy(),
+            inc_metrics=self.inc_metrics.copy(),
+            metadata=self.metadata,
+            profile=self.profile,
+            profile_mapping=self.profile_mapping,
```
ilumsden: I'd copy `self.metadata`, `self.profile`, and `self.profile_mapping` here since they shouldn't be affected by the `squash`.

For `self.metadata`, you do this with `self.metadata.copy()` (with or without `deep=True`; I don't think that really matters here).

For the other two, you can copy with `copy.copy(self.profile)` or `copy.deepcopy(self.profile)`. See Thicket's `copy` and `deepcopy` methods for examples.

ilumsden (changes requested):
I suspect that the next review will be an approval for this PR.

There are just a handful of minor things to change. Most of them amount to removing stringification of multi-level column names for the cache. The remaining changes are just minor one line things.

ilumsden (approved):
Looks good! @pearce8 this PR is ready for your review

pearce8:
@Yejashi Do we have docs/examples (tutorial?) to go with this PR?
@Yejashi @ilumsden Any further thoughts on how we could enforce that the future functions use the decorator?  Example of using this with a different decorator (e.g., profiling the functions)?

Yejashi:
@pearce8 , in terms of documentation, there really isn't any meaningful difference on how the stats functions are called. All that's really changed from the user end of things is that the stats functions now return a list of output columns. This is already indicated in the docstrings and but not in the thicket documentation website. A concise sentence could accomplish this task.

In terms of queries, i think we should add a small section in the query portion of the tutorials indicating that the query language can now be used on the statsframe, albeit with a slight difference. An example or two would be enough for this.

ilumsden:
@pearce8 I mentioned this on Slack in more detail, but this branch shows an example of using Befikir's cache decorator with other decorators: https://github.com/TauferLab/thicket/tree/memray_decorator_example

pearce8 (approved):

PR #163, opened 2024-05-13
Improve Reader Error Messages

michaelmckinsey1:
This PR improves the error messages in the readers.
- Add separate check for when empty list or tuple is provided.
- Refactor error checking to be separate from the file processing.
- Add unit tests to ensure the correct error messages are being raised.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -304,6 +304,12 @@ class Thicket(GraphFrame):
             tdmq_output (bool): whether to display tqdm progress bar
             args (list): list of args; args[0] should be an object that can be read from
         """
+
+        def check_file_exists(file):
+            if not os.path.isfile(file):
+                raise FileNotFoundError("File '" + file + "' not found")
+            return True
+
         ens_list = []
         obj = args[0]  # First arg should be readable object
         extra_args = []
@@ -314,6 +320,10 @@ class Thicket(GraphFrame):
         # Parse the input object
         # if a list of files
         if isinstance(obj, (list, tuple)):
+            if len(obj) == 0:
+                raise ValueError("Iterable must contain at least one file")
+            for file in obj:
+                check_file_exists(file)
             pbar = tqdm.tqdm(obj, disable=disable_tqdm)
             for file in pbar:
                 pbar.set_description(pbar_desc)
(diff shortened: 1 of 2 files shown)


thicket/thicket.py
```diff
@@ -304,13 +304,33 @@ def reader_dispatch(func, intersection, disable_tqdm, *args, **kwargs):
...
+                raise FileNotFoundError("File '" + file + "' not found")
+            return True
+
         ens_list = []
         obj = args[0]  # First arg should be readable object
-        extra_args = []
         if len(args) > 1:
             extra_args = args[1:]
+        else:
+            extra_args = []
```
ilumsden: This isn't a problem to do, but it will introduce a (very tiny) performance hit. In general, regardless of programming language, branches are extremely expensive, and so, the more you have, the slower your code will be. When I say "branches", I'm referring to every part of every if-elif-else statement.

thicket/thicket.py
```diff
@@ -304,13 +304,33 @@ def reader_dispatch(func, intersection, disable_tqdm, *args, **kwargs):
...
         pbar_desc = "(1/2) Reading Files"
 
+        # Error checking
+        if isinstance(obj, (list, tuple)):
+            if len(obj) == 0:
+                raise ValueError("Iterable must contain at least one file")
+            for file in obj:
+                check_file_exists(file)
+        elif isinstance(obj, str):
+            check_file_exists(obj)
```
ilumsden: Just put the if statement from `check_file_exists` in both of these places. Minimizing code duplication is useful, but, when there's only two spots where it's going to be called and there's only two lines of potentially duplicated code, it's better to just copy the code and avoid the overhead of a function call.

ilumsden (approved):
Looks good. I've got two tiny suggestions, but they're minor enough that I'll approve.

thicket/thicket.py
michaelmckinsey1: This is in `reader_dispatch` so it's ran once each time a reader is called. Therefore it is really a negligible performance hit, but actually since you pointed it out I realized:
```
if len(args) > 1:
            extra_args = args[1:]
else:
            extra_args = []
```
is the same as
```
extra_args = args[1:]
```
so I changed it anyway.

thicket/thicket.py
michaelmckinsey1: Sure. Done.

PR #168, opened 2024-06-10
Add Option to Turn Off _fill_perfdata

michaelmckinsey1:
`_fill_perfdata` adds empty rows for profiles that do not contain data for a node, such that each node has a row entry for every profile. This can be obnoxious in some cases, as in the performance algorithms dataset, 2,531,150 rows in the performance data table with `_fill_perfdata` and 163,778 rows with `_fill_perfdata` off.

This is a small optimization for the reader, as `_fill_perdata` is only called once, most time is spent in `_unify`. However, when working with a large amount of files with missing rows, a Thicket with `_fill_perfdata` off will be easier and faster to work with.

### DIFF
--- a/thicket/ensemble.py
+++ b/thicket/ensemble.py
@@ -369,12 +369,15 @@ class Ensemble:
         return combined_th
 
     @staticmethod
-    def _index(thickets, from_statsframes=False, disable_tqdm=False):
+    def _index(
+        thickets, from_statsframes=False, fill_perfdata=True, disable_tqdm=False
+    ):
         """Unify a list of thickets into a single thicket
 
         Arguments:
             thickets (list): list of Thicket objects
             from_statsframes (bool): Whether this method was invoked from from_statsframes
+            fill_perfdata (bool): whether to fill missing rows in performance data table
             disable_tqdm (bool): whether to disable tqdm progress bar
 
         Returns:
@@ -456,7 +459,8 @@ class Ensemble:
         validate_dataframe(unify_df)
 
         # Insert missing rows in dataframe
-        unify_df = _fill_perfdata(unify_df)
+        if fill_perfdata:
+            unify_df = _fill_perfdata(unify_df)
 
         # Sort PerfData
         unify_df.sort_index(inplace=True)
--- a/thicket/ensemble.py
+++ b/thicket/ensemble.py
@@ -377,7 +377,7 @@ class Ensemble:
         Arguments:
             thickets (list): list of Thicket objects
             from_statsframes (bool): Whether this method was invoked from from_statsframes
-            fill_perfdata (bool): whether to fill missing rows in performance data table
+            fill_perfdata (bool): whether to fill missing performance data with NaNs
             disable_tqdm (bool): whether to disable tqdm progress bar
 
         Returns:
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -203,7 +203,11 @@ class Thicket(GraphFrame):
 
     @staticmethod
     def from_caliper(
-        filename_or_stream, query=None, intersection=False, disable_tqdm=False
+        filename_or_stream,
+        query=None,
+        intersection=False,
+        fill_perfdata=True,
+        disable_tqdm=False,
     ):
         """Read in a Caliper .cali or .json file.
 
@@ -212,36 +216,48 @@ class Thicket(GraphFrame):
                 `.cali` or JSON-split format, or an open file object to read one
             query (str): cali-query in CalQL format
             intersection (bool): whether to perform intersection or union (default)
+            fill_perfdata (bool): whether to fill missing performance data with NaNs
             disable_tqdm (bool): whether to display tqdm progress bar
         """
         return Thicket.reader_dispatch(
             GraphFrame.from_caliper,
             intersection,
+            fill_perfdata,
             disable_tqdm,
             filename_or_stream,
             query,
         )
 
     @staticmethod
-    def from_hpctoolkit(dirname, intersection=False, disable_tqdm=False):
+    def from_hpctoolkit(
+        dirname, intersection=False, fill_perfdata=True, disable_tqdm=False
+    ):
         """Create a GraphFrame using hatchet's HPCToolkit reader and use its attributes
         to make a new thicket.
 
         Arguments:
             dirname (str): parent directory of an HPCToolkit experiment.xml file
             intersection (bool): whether to perform intersection or union (default)
+            fill_perfdata (bool): whether to fill missing performance data with NaNs
             disable_tqdm (bool): whether to display tqdm progress bar
 
         Returns:
             (thicket): new thicket containing HPCToolkit profile data
         """
         return Thicket.reader_dispatch(
-            GraphFrame.from_hpctoolkit, intersection, disable_tqdm, dirname
+            GraphFrame.from_hpctoolkit,
+            intersection,
+            fill_perfdata,
+            disable_tqdm,
+            dirname,
         )
 
     @staticmethod
     def from_caliperreader(
-        filename_or_caliperreader, intersection=False, disable_tqdm=False
+        filename_or_caliperreader,
+        intersection=False,
+        fill_perfdata=True,
+        disable_tqdm=False,
     ):
         """Helper function to read one caliper file.
 
@@ -249,11 +265,13 @@ class Thicket(GraphFrame):
             filename_or_caliperreader (str or CaliperReader): name of a Caliper output
                 file in `.cali` format, or a CaliperReader object
             intersection (bool): whether to perform intersection or union (default)
+            fill_perfdata (bool): whether to fill missing performance data with NaNs
             disable_tqdm (bool): whether to display tqdm progress bar
         """
         return Thicket.reader_dispatch(
             GraphFrame.from_caliperreader,
             intersection,
+            fill_perfdata,
             disable_tqdm,
             filename_or_caliperreader,
         )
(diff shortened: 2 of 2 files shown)


michaelmckinsey1:
Need to add unit test.

thicket/ensemble.py
```diff
@@ -456,7 +459,8 @@ def _fill_perfdata(df, numerical_fill_value=np.nan):
         validate_dataframe(unify_df)
 
         # Insert missing rows in dataframe
-        unify_df = _fill_perfdata(unify_df)
+        if fill_perfdata:
+            unify_df = _fill_perfdata(unify_df)
```
ilumsden, line 463: This looks good here. However, I took a second to look at the implementation of `_fill_perfdata` (I didn't realize this was a thing before), and I do see an issue there. In `_fill_perfdata`, there's the following code:
```python
except ValueError as e:
    estr = str(e)
    if estr == "cannot handle a non-unique multi-index!":
        warnings.warn(
            "Non-unique multi-index for DataFrame in _fill_perfdata. Cannot Fill missing rows.",
            RuntimeWarning,
        )
    else:
        raise
```

You should **never** depend on another library's error message to report warnings or errors. All it will take to make this code completely invalid is someone from the `pandas` team deciding to change even one character of this message. Either find another way to check when to throw or warn, or always throw a slightly less descriptive error.

thicket/thicket.py
```diff
@@ -1112,6 +1140,12 @@ def filter_metadata(self, select_function):
         else:
             raise InvalidFilter("The argument passed to filter must be a callable.")
 
+        # If fill_perfdata is False, may need to squash
+        unique_perf_indices = set(new_thicket.dataframe.index.droplevel("node"))
```
ilumsden: Can you cast a pandas `Index` or `MultiIndex` to a set? This seems like an invalid operation that, at best, is creating a set of size 1 containing the pandas object.

thicket/thicket.py
```diff
@@ -1112,6 +1140,12 @@ def filter_metadata(self, select_function):
...
+        full_idx = all([idx in unique_perf_indices for idx in new_thicket.profile])
+        if not full_idx:
+            new_thicket = new_thicket.squash()
```
ilumsden, line 1147: I'm not sure what this is trying to do. From what I can tell, this is trying to see if the `profile` list is a subset of the unique profile IDs in the performance dataframe index. If that's true, then the dataframe can store profile IDs that aren't in the `Thicket.profile` field.

ilumsden (commented):
I have a few comments/changes, but this mostly looks good.

The one thing I'd mention is that this **will** have a notable impact in performance in some cases. As you said in the initial PR comment, this functionality can substantially increase the number of rows in the performance dataframe. Given how slow pandas is and given that pandas uses a minimum of 2x the memory that a single `DataFrame` _should_ take up, adding a substantial number of nodes **will** slow down **all** of Thicket.

~Given that, my one remaining comment (which we'll need to discuss) is that it might be better to keep this off by default. Right now in this PR, users have to opt *out* of this feature, which could destroy Thicket's performance if used in the wrong situation. I'd argue that it's better for users to opt *into* this feature in situations where they know the performance won't be destroyed.~ Addressed in the discussion below. This is the default behavior of Thicket right now, so this PR shouldn't change that. However, IMO, this will cause issues in the future.

ilumsden (changes requested):
Accidentally forget to set "Request changes" on my main review. Refer to the main review for comments.

michaelmckinsey1:
> I have a few comments/changes, but this mostly looks good.
> 
> The one thing I'd mention is that this **will** have a notable impact in performance in some cases. As you said in the initial PR comment, this functionality can substantially increase the number of rows in the performance dataframe. Given how slow pandas is and given that pandas uses a minimum of 2x the memory that a single `DataFrame` _should_ take up, adding a substantial number of nodes **will** slow down **all** of Thicket.
> 
> Given that, my one remaining comment (which we'll need to discuss) is that it might be better to keep this off by default. Right now in this PR, users have to opt _out_ of this feature, which could destroy Thicket's performance if used in the wrong situation. I'd argue that it's better for users to opt _into_ this feature in situations where they know the performance won't be destroyed.

It's on by default because the default behavior in Thicket has been to fill the index in the performance table. Actually, the default behavior _originally_ was to not fill the index, then there was a design decision made at some point to fill it all the time. My goal with this PR is to at least give the user the option, based on their data.

ilumsden:
Wait, really? In that case, what you have in this PR is fine, but I will second guess that decision. I suspect we haven't run into any major performance issues because we've just used effectively _N_ copies of the same program. But, once we run into less homogeneous data (e.g., MPMD-style data like what my benchmark will create), we will likely start running into major performance issues with this decision.

thicket/thicket.py
michaelmckinsey1: This is detecting if `fill_perfdata` is off. To make it more clear I thought of a couple options:

1. I could make it a helper function with a docstring to make it more clear.
2. Have a boolean thicket attribute for `fill_perfdata` that can be checked here.

The purpose of this is that there is a case when the index is not full, and a filter is performed on the metadata that nodes disappear from the performance data, if the metadata for those profiles was filtered out. The code above will update the graph using squash if this happens.

thicket/thicket.py
michaelmckinsey1: Yes you can.

image

The set of the index is all of the unique values in the index.

thicket/thicket.py
ilumsden: So, if I'm understanding right, what you're trying to do is 1) check if any nodes were removed from the performance data as a side effect of filtering profiles from the metadata and 2) fix the graph with `squash` if any nodes were removed?

ilumsden: If what I wrote above is correct, then this code will not do what you're wanting.

ilumsden: This code (assuming the `set(Series)` thing works) only checks if `Thicket.profiles` is a subset of the profile IDs in the performance dataframe index. That tells you nothing about the nodes.

michaelmckinsey1: It tells you if there are profiles in the performance dataframe that do not exist in `tk.profile`. For example if I simulate removing a profile from the dataframe, it will be false which will trigger a squash:

image

The actual issue I noticed with this is that they will both always be in sync due to `_sync_profile_components`. So what I should check here instead is if `len(tk.graph) != len(tk.dataframe.index.get_level_values("node").unique())` which is more explicit anyway. These sort of things will get caught when I add unit tests.

thicket/ensemble.py
michaelmckinsey1: I made an issue for this https://github.com/LLNL/thicket/issues/172.

ilumsden: Sounds good

thicket/thicket.py
ilumsden: Huh. I wouldn't have expected that to work. Pandas usually gets very finicky when applying normal Python operations to `Series`. This works for me.

thicket/thicket.py
ilumsden: Your code is not checking if there are profiles in the dataframe that don't exist in `tk.profile`. It's checking if there are profiles in `tk.profile` that don't exist in the dataframe. It's a subtle difference, but it would allow e.g., the dataframe to contain extra profiles that don't exist in `tk.profiles`.

I actually just encountered this same type of thing in Befikir's `query_stats` PR (#157), and I realized there's a better way to do this check:
```python
sorted(tk.profile) == sorted(unique_perf_indices.tolist())
```

This will check that both the dataframe and `tk.profile` agree in terms of contents and length. `Node` objects are comparable via their `_hatchet_nid` fields. As a result, the `Node` objects within different parts *of the same `Thicket` object* can be compared in this way. This comparison will **not** work across `Thicket` objects though because there's no guarantee that equivalent `Node` objects will have the same `_hatchet_nid` values.

michaelmckinsey1: > It's checking if there are profiles in tk.profile that don't exist in the dataframe.

But that is what I was trying to accomplish with that code. That would indicate that there are missing profiles in the performance data, which is what I was saying. However, it doesn't address 

> it would allow e.g., the dataframe to contain extra profiles that don't exist in tk.profiles.

which your correct would be an issue, if not for `_sync_profile_components` being called earlier. Regardless, what I was actually trying to do is check nodes so I'm using a different check now.

michaelmckinsey1: For reference, if we are trying to update the state of the profiles in a Thicket we should be using `_sync_profile_components`. And if we are trying to validate properties about the state of a Thicket we should use `validate_profile` or `validate_dataframe` in `utils.py`

ilumsden: If you're 100% positive that you've ensured the graph and dataframe contain the same node objects, then this is fine.

ilumsden (approved):
Looks good, so long as we are positive that the `Node` objects are the same across thicket components.

michaelmckinsey1:
> Looks good, so long as we are positive that the `Node` objects are the same across thicket components.

They should be unless there is a bug. `utils.validate_nodes` is a function that checks this particular property. If we wanted to be 100% sure, I think we would have to check properties we want to enforce about a Thicket before and after operations like `filter_metadata`, which could be expensive. I am open to this idea though, as it would help catch bugs.

Edit: I just realized we had a basically identical discussion in https://github.com/LLNL/thicket/pull/173#discussion_r1643134992 so I am referencing it here. I agree that this is a broader issue than both of these PRs.

michaelmckinsey1:
> > Looks good, so long as we are positive that the `Node` objects are the same across thicket components.
> 
> They should be unless there is a bug. `utils.validate_nodes` is a function that checks this particular property. If we wanted to be 100% sure, I think we would have to check properties we want to enforce about a Thicket before and after operations like `filter_metadata`, which could be expensive. I am open to this idea though, as it would help catch bugs.

I just realized we had a basically identical discussion in https://github.com/LLNL/thicket/pull/173#discussion_r1643134992. I agree that this is a broader issue than both of these PRs.

pearce8 (approved):

PR #169, opened 2024-06-10
Add Option to Speedup Thicket Reader with Multiprocess

michaelmckinsey1:
Add optional `parallel` flag in readers to use the `multiprocessing` library to compose Thickets in parallel. For a lot of files, composing Thickets takes a majority of the time over reading, and the rate of composition slows down as the number of files increases. 

Our process right now is:
1. Read single-profile thickets using Hatchet readers into a list.
2. Compose the list of Thickets into one big Thicket.

We can divide the list in step 2 into chunks and do the step in `n` parallel chunks. Once it is done we can compose the remaining `n` thickets.

An example of how this helps (parallel algorithms dataset):
```
sequential: 16,330 files takes 120 minutes
parallel (n=16): 16,330 files takes 3 minutes
```
40x speedup is possible with `n=16` because as stated above, the rate of composition slows down as files increase, due to the nature of the unification process. So by composing in parallel, we never reach as slow of a composition rate.

Update - I tested splitting the files up without the parallelism, showing most of the work is just splitting up the max file count:
```
sequential (16 parts): 16,330 files takes 6 minutes
```

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -4,12 +4,14 @@
 # SPDX-License-Identifier: MIT
 
 import copy
+import multiprocessing
 import os
 import sys
 import json
 import warnings
 from collections import defaultdict, OrderedDict
 from hashlib import md5
+from functools import partial
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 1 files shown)


michaelmckinsey1:
Holding off on using `multiprocessing` for now in favor of the approach in #170

PR #170, opened 2024-06-11
Speedup thicket composition by performing profile unions in binary tree order

michaelmckinsey1:
This PR extends off #169. Limiting the total size of the ensemble during unification is more impactful than the performance gains from parallelism, and this solution is less complicated than using `multiprocessing`.

| | Parallel Sorting | LBANN |
|---|---|---|
| Profiles | 15734 | 16 |
| Input Nodes (min, max) | 5, 18 | 6266, 6889 |
| Input Dataframe Size (min, max) | '' | '' |
| Output Nodes | 154 | 7106 |
| Output DF | 2423036 | 113696 |

| Dataset | Thicket Version | Time | Unions | old_to_new size (min, max) |
|---|---|---|---|---|
| Parallel Sorting | Develop | >120m | 15733 | 157459 |
|  | #170 | 12m | '' | 16, 185 |
|  | #185 | 33m | '' | 157459 |
| LBANN | Develop | 17m | 15 | 106897 |
| | #170 | 7m | '' | 12824, 13821 |
| | #185 | 3m | '' | 106897 |

Number of unions is not affected by the different unifying strategy in this PR. This is because:
```
# PR170 - 3 unions
tk1 U tk2 = tk12
tk3 U tk4 = tk34
tk12 U tk34 = tk1234

is the same as

# develop - 3 unions
tk1 U tk2 = tk12
tk12 U tk3 = tk123
tk123 U tk4 = tk1234
```
This PR increases performance by keeping the `old_to_new` mapping dictionary small as we call `concat_thickets` `n-1` times instead of one time (as in the develop branch). The `old_to_new` dictionary requires an operation to update the dictionary mappings per each union, which can become expensive if the dictionary is too large. The `old_to_new` dictionary is necessary to replace the node objects in the performance data with the new node objects in the graph that results from `union`.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -390,11 +390,13 @@ class Thicket(GraphFrame):
         calltree = "union"
         if intersection:
             calltree = "intersection"
-        # We concatenate t Thickets with 1 prof each until 1 Thicket with t profs.
         # Consider a binary tree with t leaves (Thickets) => t-1 parents (Knuth 1997).
         # Each non-leaf Thicket (parent) is a result of a concat_thickets, therefore
         # we have t-1 concat_thickets operations. t == len(ens_list)
         pbar = tqdm.tqdm(total=len(ens_list) - 1, disable=disable_tqdm)
+        # We start with t Thickets with 1 prof each until 1 Thicket with t profs.
+        # Each iteration is concatenating 2 Thickets with n, m profs into 1 Thicket 
+        # with n+m profs, appending the new Thicket to the back of the ens_list.
         while len(ens_list) > 1:
             pbar.set_description("(2/2) Creating Thicket")
             new_tk = Thicket.concat_thickets(
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 import copy
+import math
 import os
 import pickle
 import sys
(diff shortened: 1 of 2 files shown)


thicket/thicket.py
```diff
@@ -349,6 +350,21 @@ def reader_dispatch(func, intersection, disable_tqdm, *args, **kwargs):
         calltree = "union"
         if intersection:
             calltree = "intersection"
+
+        num_tks = len(ens_list)
+        tk_limit = 2**9
```
michaelmckinsey1: need to figure out a good way to set this value

thicket/thicket.py
```diff
@@ -353,14 +353,19 @@ def reader_dispatch(func, intersection, disable_tqdm, *args, **kwargs):
...
-        thicket_object = Thicket.concat_thickets(
-            thickets=ens_list,
-            axis="index",
-            calltree=calltree,
-            disable_tqdm=disable_tqdm,
-        )
-
-        return thicket_object
+        # n - 1 edges in a binary tree
+        pbar = tqdm.tqdm(range(len(ens_list) - 1), disable=disable_tqdm)
```
ilumsden: Using `tqdm` here makes this process **very** hard to understand. I'd suggest not using `tqdm` and/or extensively documenting this with comments.

ilumsden (changes requested):
The code is good, but something has to be done to make the tree-based concatenation clearer. At bare minimum, this needs to be well commented, otherwise no one is going to have a clue what this is doing.

thicket/thicket.py
michaelmckinsey1: `tqdm` is really useful to see what is happening for longer reads.

I changed it to use a while loop, which should be clearer with the popping and appending that is happening in the loop. And I added very detailed comments.

ilumsden (approved):
LGTM. The stuff you're doing with tqdm also makes much more sense now that you've added the comments and switched to a `while` loop

ilumsden:
@pearce8 this PR is also ready for your review

pearce8 (approved):

PR #173, opened 2024-06-12
Enable Filtering From List of Profiles

michaelmckinsey1:
I have came across the need to filter the Thicket based on profiles, and I have been using a roundabout way:
```
# Filter the metadata table based on list of profiles
tk.metadata = tk.metadata.apply(lambda meta: meta[meta.index.isin(profile_list)], axis=0)
# Synchronize the rest of the components
tk._sync_profile_components(tk.metadata)
```
So I designed an explicit function for this filtering `filter_profile`.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1207,7 +1207,7 @@ class Thicket(GraphFrame):
         Returns:
             (thicket): new thicket object with selected profiles
         """
-        new_thicket = self.copy()
+        new_thicket = self.deepcopy()
 
         new_thicket._sync_profile_components(profile_list)
         validate_profile(new_thicket)
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1496,9 +1496,6 @@ class Thicket(GraphFrame):
         Arguments:
             component (list or DataFrame) -> (list, Thicket.dataframe, or Thicket.metadata): The index
             of this component is used to synchronize the other objects.
-
-        Returns:
-            (thicket): self
         """
 
         def _profile_truth_from_component(component):
@@ -1563,8 +1560,6 @@ class Thicket(GraphFrame):
             raise ValueError(
                 "Component must be either list, Thicket.dataframe, or Thicket.metadata"
             )
-
-        return self
 
 
 class InvalidFilter(Exception):
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1550,10 +1550,10 @@ class Thicket(GraphFrame):
             return self
 
         if isinstance(component, list):
-            self = _sync_indices(component, component)
+            self = _sync_indices(component)
         elif isinstance(component, pd.DataFrame):
             profile_truth = _profile_truth_from_component(component)
-            self = _sync_indices(component, profile_truth)
+            self = _sync_indices(profile_truth)
         else:
             raise ValueError(
                 "Component must be either list, Thicket.dataframe, or Thicket.metadata"
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1215,7 +1215,7 @@ class Thicket(GraphFrame):
         if len(new_thicket.graph) != len(
             new_thicket.dataframe.index.get_level_values("node").unique()
         ):
-            new_thicket.squash()
+            new_thicket = new_thicket.squash()
 
         return new_thicket
 
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1523,7 +1523,7 @@ class Thicket(GraphFrame):
                     profile_truth = component.index
             return list(set(profile_truth))
 
-        def _sync_indices(component, profile_truth):
+        def _sync_indices(profile_truth):
             """Sync the Thicket attributes"""
             self.profile = profile_truth
             self.profile_mapping = OrderedDict(
@@ -1534,10 +1534,8 @@ class Thicket(GraphFrame):
                 }
             )
 
-            if isinstance(component, list):
-                pass
             # For Columnar-indexed Thicket
-            elif isinstance(component.columns, pd.MultiIndex):
+            if isinstance(self.dataframe.columns, pd.MultiIndex):
                 # Create powerset from all profiles
                 pset = set()
                 for p in profile_truth:
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1198,6 +1198,27 @@ class Thicket(GraphFrame):
 
         return new_thicket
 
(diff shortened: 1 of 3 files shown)


thicket/thicket.py
```diff
@@ -1114,6 +1114,27 @@ def filter_metadata(self, select_function):
...
+    def filter_profile(self, profile_list):
+        """Filter thicket object based on a list of profiles.
+
+        Arguments:
+            profile_list (list): list of profiles to filter on
+
+        Returns:
+            (thicket): new thicket object with selected profiles
+        """
+        new_thicket = self.copy()
```
ilumsden: Should this be `copy()` or `deepcopy()`? In the case of `copy()`, the graph is shared between the input Thicket (i.e., `self`) and `new_thicket`. So, if either Thicket object edits its graph, the change will be reflected in the other Thicket.

thicket/thicket.py
```diff
@@ -1114,6 +1114,27 @@ def filter_metadata(self, select_function):
...
+
+        new_thicket._sync_profile_components(profile_list)
+        validate_profile(new_thicket)
+
+        if len(new_thicket.graph) != len(
+            new_thicket.dataframe.index.get_level_values("node").unique()
```
ilumsden, line 1216: Are you 100% positive that the actual `Node` objects in `new_thicket.graph` and `new_thicket.dataframe` are the same here? If not, then this check is invalid because it only sees if there's the same number of nodes, not if the nodes themselves are the same.

An alternative to check for both would be:
```python
sorted(new_thicket.graph.traverse()) == sorted(new_thicket.dataframe.index.get_level_values("node").unique().tolist())
```

This alternative checks that the number of nodes are the same **and** that the nodes themselves are the same.

thicket/thicket.py
```diff
@@ -1345,8 +1366,10 @@ def _sync_indices(component, profile_truth):
                 }
             )
 
+            if isinstance(component, list):
+                pass
             # For Columnar-indexed Thicket
-            if isinstance(component.columns, pd.MultiIndex):
+            elif isinstance(component.columns, pd.MultiIndex):
                 # Create powerset from all profiles
                 pset = set()
```
ilumsden, line 1542: I get why this is a set here, but cast it to a `list` or `tuple` before assigning to `profile_truth` for consistency. You could potentially run into issues having `profile_truth` be a `list` sometimes and a `set` others.

thicket/thicket.py
```diff
@@ -1360,14 +1383,16 @@ def _sync_indices(component, profile_truth):
...
+        else:
             raise ValueError(
-                "Component must be either Thicket.dataframe or Thicket.metadata"
+                "Component must be either list, Thicket.dataframe, or Thicket.metadata"
             )
 
-        profile_truth = _profile_truth_from_component(component)
-        self = _sync_indices(component, profile_truth)
-
         return self
```
ilumsden: Why do you return `self` here? Is it for method chaining? If not, I don't really get the purpose of doing this.

ilumsden (changes requested):
Mostly looks good. Just a few minor changes.

thicket/thicket.py
michaelmckinsey1: Good catch. Objects should not be shared between the two Thickets. I realized this same issue was in the other filter functions so I made https://github.com/LLNL/thicket/pull/175

thicket/thicket.py
michaelmckinsey1: This function will not modify the `Node` objects in a Thicket `graph` and `dataframe`. The only possible case is that `_sync_profile_components` would _remove_ nodes in the dataframe where `fill_perfdata=False`. If the `Node` objects in a Thicket `graph` and `dataframe` are not the same that would be an invalid state for a Thicket coming into this function (which `utils.validate_nodes` would be able to catch, although we are not calling it here).

So the short answer is what you are proposing is not possible unless the Thicket calling `filter_profile` is already invalid. Right now, the utils functions `validate_dataframe`, `validate_profile`, and `validate_nodes` are only called after operations applicable to each one for performance, but maybe we should be validating these properties at the beginning and end of Thicket funcitons.

thicket/thicket.py
michaelmckinsey1: Python is magical! Fixed in [6bf7ce8](https://github.com/LLNL/thicket/pull/173/commits/6bf7ce824458638297c0fcac9590a65397e00810)

thicket/thicket.py
michaelmckinsey1: There is no specific reason. I think it could be void. Is there a specific problem you had in mind or just for clarity?

thicket/thicket.py
ilumsden: I get your point, and I agree that checking this when not needed would hurt performance. I think this is a broader discussion than we need for this PR. We should come back to this at some point, but this is fine for this PR.

thicket/thicket.py
ilumsden: In general, especially given that Python is duck typed, it's better to not return stuff if it's not supposed to be used by the user. So, for example, returning `self` should only be done if you intend for the user to use that return value for e.g., method chaining. If you start returning stuff when you don't have to, it gets really confusing for users really fast.

thicket/thicket.py
ilumsden: Awesome

ilumsden (approved):
I'd still remove the `return self` (as I mentioned in the last review), but otherwise this looks good

pearce8 (approved):

pearce8 (changes requested):
Are there docs/test/examples for this functionality?

thicket/thicket.py
pearce8: Lets change it to void then.

michaelmckinsey1:
> Are there docs/test/examples for this functionality?

No. This PR needs unit tests.

thicket/thicket.py
michaelmckinsey1: Done in [1758146](https://github.com/LLNL/thicket/pull/173/commits/1758146dbe57c3ff6011eda391d29a9765428543)

michaelmckinsey1:
> > Are there docs/test/examples for this functionality?
> 
> No. This PR needs unit tests.

Unit tests in [f0b2f50](https://github.com/LLNL/thicket/pull/173/commits/f0b2f50ec965a45f3f205a1c56f541a4a057fe7c)

docs https://github.com/LLNL/thicket-tutorial/pull/47

pearce8 (approved):

PR #174, opened 2024-06-13
print thicket tree by node order (hatchet nid) instead of alphabetically

dyokelson:
This changes the printing of the thicket tree from alphabetical (frame) to node order (_hatchet_nid). If the profile does not have node ordering (from old caliper, or some other format) it will still sort the print by _hatchet_nid, which can be assigned alphabetically, possibly resulting in alphabetical printing of the thicket tree.


thicket/tests/test_caliperreader.py
```diff
@@ -23,3 +23,66 @@ def test_from_caliperreader(rajaperf_seq_O3_1M_cali):
...
+        node_name = tk.dataframe.iloc[i]["name"]
+        assert node_name == expected_order[i]
+        node_data = tk.dataframe.iloc[i]["Total time"]
+        assert node_data == expected_data_order[i]
+
+    # check the tree ordering is correct as well
+    output = tk.tree(metric_column="nid")
+    for i in range(1, 20):
+        location = output.find(str(i))
+        assert location != 0
```
michaelmckinsey1:
```suggestion
    tk.dataframe["hnid"] = [node._hatchet_nid for node in tk.graph.node_order_traverse()]
    output = tk.tree(metric_column="hnid", render_header=False, precision=3)
    for i in tk.dataframe["hnid"].tolist():
        location = output.find(str(i)+".000")
        assert location != -1
```
A couple of things I noticed in this check:

1. The `nid` column is not the same as the `_hatchet_nid`, so to be explicit we should check for the `_hatchet_nid`. In this case the `nid` = `_hatchet_nid + 1`, I'm not sure if there's a case where they would be in different orders.
2. Render the tree with `hnid`, set the header to false, and set the precision to 3 for the find function. 3 is the default but just in case the default changes.
3. check `location != -1` instead of `0` as `find` will return `-1` if the value is not found.

michaelmckinsey1 (changes requested):

thicket/tests/test_caliperreader.py
dyokelson: Good idea, I think we should be as explicit as possible.

pearce8 (approved):

michaelmckinsey1 (approved):

PR #175, opened 2024-06-13
Change Filter Shallow Copies to Deep Copies

michaelmckinsey1:
@ilumsden made a good catch in https://github.com/LLNL/thicket/pull/173, that filter copies should be deep. This PR changes `filter_metadata` and `filter_stats` shallow copies to be deep copies. This is because we expect the returned Thicket from a filter to be a separate Thicket object.

Example of changing the filtered Thicket:
image
Shallow Copy - both objects change
image
Deep Copy - only `th` changes
image

For the sake of a unit test it is sufficient to check that a given filtered Thicket component `is not` the same object as the component in the original Thicket.

### DIFF
--- a/thicket/tests/test_filter_metadata.py
+++ b/thicket/tests/test_filter_metadata.py
@@ -32,6 +32,9 @@ def filter_one_column(th, columns_values):
             # check if output is a thicket object
             assert isinstance(new_th, Thicket)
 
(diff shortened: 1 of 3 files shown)


ilumsden (approved):
Looks good to me. @pearce8 this PR is ready for your review

PR #176, opened 2024-06-14
docs: built-in configuration page link extension

guspan-tanadi:
A link in Generating Data page having extension as `html` as followed by other link.


ilumsden (approved):
LGTM.

@pearce8 this is a minor URL fix for our docs, so I'd argue that we can go ahead and merge this before Stephanie's back.

PR #178, opened 2024-06-18
Check if Metric Exists in add_ncu

michaelmckinsey1:
@Yejashi discovered that running `add_ncu` twice will cause the metrics to be added twice with `_left` and `_right` suffixes, as expected from the `join` called at the end of `add_ncu`. However, this may be unintuitive for our users, so this PR adds a check if the column already exists taking inspiration from the check in `metadata_column_to_perfdata`. The user has the option to overwrite the columns by setting `overwrite=True`. ~This PR also refactors the check in `metadata_column_to_perfdata` to a `utils` function so it can be used by `metadata_column_to_perfdata`, `add_ncu`, and any other functions that require this check.~

### DIFF
--- a/thicket/utils.py
+++ b/thicket/utils.py
@@ -63,6 +63,25 @@ def check_duplicate_metadata_key(thickets, metadata_key):
         )
 
 
+def overwrite_column(df, col, overwrite):
+    """Check and overwrite a column in a DataFrame if it already exists.
+
+    Arguments:
+        df (DataFrame): Input DataFrame
+        col (str): Column name to check
+        overwrite (bool): Option to overwrite column if it already exists
+
+    Returns:
+        (DataFrame): DataFrame with column removed
+    """
+    if overwrite:
+        return df.drop(col, axis=1)
+    else:
+        raise ValueError(
+            f"Column {col} already exists. Set 'overwrite=True' to overwrite."
+        )
+
+
 def validate_dataframe(df):
     """Check validity of a Thicket DataFrame."""
 
--- a/thicket/utils.py
+++ b/thicket/utils.py
@@ -63,25 +63,6 @@ def check_duplicate_metadata_key(thickets, metadata_key):
         )
 
 
-def overwrite_column(df, col, overwrite):
-    """Check and overwrite a column in a DataFrame if it already exists.
-
-    Arguments:
-        df (DataFrame): Input DataFrame
-        col (str): Column name to check
-        overwrite (bool): Option to overwrite column if it already exists
-
-    Returns:
-        (DataFrame): DataFrame with column removed
-    """
-    if overwrite:
-        return df.drop(col, axis=1)
-    else:
-        raise ValueError(
-            f"Column {col} already exists. Set 'overwrite=True' to overwrite."
-        )
-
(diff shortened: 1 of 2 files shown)


thicket/utils.py
```diff
@@ -63,6 +63,25 @@ def check_duplicate_metadata_key(thickets, metadata_key):
         )
 
 
+def overwrite_column(df, col, overwrite):
```
ilumsden: Minor nitpick, but I think this function should have a different name. This function is **not** overwriting a column. It is dropping a column, so that it can be (effectively) overwritten. An actual `overwrite_column` function should essentially do a drop + an assign for the new version of the column.

ilumsden: Actually, another thing.

If you're calling a function to overwrite (or really drop) a column, why would you have a boolean telling that function whether or not to override? That defeats the purpose of the entire function.

thicket/utils.py
```diff
@@ -63,6 +63,25 @@ def check_duplicate_metadata_key(thickets, metadata_key):
...
+        overwrite (bool): Option to overwrite column if it already exists
+
+    Returns:
+        (DataFrame): DataFrame with column removed
+    """
+    if overwrite:
+        return df.drop(col, axis=1)
+    else:
+        raise ValueError(
+            f"Column {col} already exists. Set 'overwrite=True' to overwrite."
```
ilumsden: You're not actually checking if the column exists though. You're just assuming that whoever passed the `overwrite` argument would only set `overwrite` to `False` if they know the column exists.

ilumsden (changes requested):
The changes to `add_ncu` look good.

However, I don't think we need the `overwrite_column` function, at least in its current form. Right now, `overwrite_column` is really a function for dropping a column, but only if the `overwrite` argument is `True`. If you're going to call this function, you're already expressing your intent to drop, so the `overwrite` argument is really behaving like one of those "Are you really sure?" prompts you see on some websites.

Given this, I can see one of 3 routes you can go with `overwrite_column`:
1. Remove `overwrite_column` as a standalone function and just incorporate the logic into `add_ncu`
2. Rename `overwrite_column` to something that indicates that it's actually just dropping a column. Move logic for checking if a column exists into this function. Remove the `overwrite` argument since the very act of calling the function indicates your intent to remove
3. Add logic to actually overwrite the column with the contents of another `pandas.Series`. In this case, you may or may not want to keep the `overwrite` argument

michaelmckinsey1:
@ilumsden The overwrite function wasn't the purpose of this PR. It was to just deduplicate the same process that is happening in `metadata_column_to_perfdata`. Ideally the overwrite would also overwrite the value, but in the case of `add_ncu` this would require running the NCU reader first (which can take minutes), so it would probably be annoying for the user to get an error after that. Maybe it was unclear but the overwrite function was not meant for users, so the check with the overwrite flag was to be passed from the parent function to tell when to raise an error.

To make this PR simpler I updated the code to have no function, although the check is basically the same that is in `metadata_column_to_perfdata`

ilumsden (approved):
LGTM

pearce8 (approved):

PR #180, opened 2024-06-20
Remove Print in Groupby

michaelmckinsey1:
There's a print in `Thicket.groupby()` that really shouldn't be there. It prints the amount of Thickets created and the dictionary of key to Thicket object.

Reasons why it should be removed:
1. pandas groupby doesn't do this.
2. The resulting `GroupBy` object can be printed after it's created anyway, as our `GroupBy` class inherits from `dict`

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1373,9 +1373,6 @@ class Thicket(GraphFrame):
                 "The provided Thicket object has an empty metadata table."
             )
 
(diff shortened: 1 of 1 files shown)


pearce8 (approved):

pearce8 (changes requested):
Fix unit tests

pearce8 (approved):

PR #181, opened 2024-06-24
Add Nodes in Slice Before Printing Tree

michaelmckinsey1:
This PR applies when `fill_perfdata` is off in the Thicket constructor, since some profiles may be missing certain nodes. This PR inserts NaNs for missing nodes, so we can still print the tree.
- Allows tree to be printed when `fill_perfdata=False`
- Moves the `_fill_perfdata` function to `utils.py` so it can be called in `tree`

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -936,28 +936,17 @@ class Thicket(GraphFrame):
         # Slices the DataFrame to simulate a single-level index
         try:
             # Select only columns used by tree for efficiency in _fill_perfdata
-            if isinstance(metric_column, list):
-                df_cols = [
-                    col
-                    for col in [
-                        *metric_column,
-                        annotation_column,
-                        name_column,
-                        context_column,
-                    ]
-                    if col in self.dataframe.columns
+            tree_cols = (
+                metric_column
+                if isinstance(metric_column, list)
+                else [metric_column]
+                + [
+                    annotation_column,
+                    name_column,
+                    context_column,
                 ]
-            else:
-                df_cols = [
-                    col
-                    for col in [
-                        metric_column,
-                        annotation_column,
-                        name_column,
-                        context_column,
-                    ]
-                    if col in self.dataframe.columns
-                ]
+            )
+            df_cols = [col for col in tree_cols if col in self.dataframe.columns]
             slice_df = self.dataframe.loc[:, df_cols]
             # _fill_perfdata to make sure number of nodes in df == graph
             slice_df = _fill_perfdata(slice_df)
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -28,6 +28,7 @@ from thicket.query import (
 import tqdm
 
 from thicket.ensemble import Ensemble
+from thicket.utils import _fill_perfdata
 
 try:
     from .ncu import NCUReader
@@ -934,8 +935,10 @@ class Thicket(GraphFrame):
         }
         # Slices the DataFrame to simulate a single-level index
         try:
+            # _fill_perfdata to make sure number of nodes in df == graph
+            slice_df = _fill_perfdata(self.dataframe)
             slice_df = (
-                self.dataframe.loc[(slice(None),) + indices, :]
+                slice_df.loc[(slice(None),) + indices, :]
                 .reset_index()
                 .set_index("node")
             )
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -935,8 +935,32 @@ class Thicket(GraphFrame):
         }
         # Slices the DataFrame to simulate a single-level index
         try:
+            # Select only columns used by tree for efficiency in _fill_perfdata
+            if isinstance(metric_column, list):
+                df_cols = [
+                    col
+                    for col in [
+                        *metric_column,
+                        annotation_column,
+                        name_column,
+                        context_column,
+                    ]
+                    if col in self.dataframe.columns
+                ]
+            else:
+                df_cols = [
+                    col
+                    for col in [
+                        metric_column,
+                        annotation_column,
+                        name_column,
+                        context_column,
+                    ]
+                    if col in self.dataframe.columns
+                ]
+            slice_df = self.dataframe.loc[:, df_cols]
             # _fill_perfdata to make sure number of nodes in df == graph
-            slice_df = _fill_perfdata(self.dataframe)
+            slice_df = _fill_perfdata(slice_df)
             slice_df = (
                 slice_df.loc[(slice(None),) + indices, :]
                 .reset_index()
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -936,17 +936,20 @@ class Thicket(GraphFrame):
         # Slices the DataFrame to simulate a single-level index
         try:
             # Select only columns used by tree for efficiency in _fill_perfdata
-            tree_cols = (
-                metric_column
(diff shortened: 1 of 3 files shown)


michaelmckinsey1:
Error found from #182

thicket/thicket.py
```diff
@@ -912,8 +913,10 @@ def tree(
         }
         # Slices the DataFrame to simulate a single-level index
         try:
+            # _fill_perfdata to make sure number of nodes in df == graph
+            slice_df = _fill_perfdata(self.dataframe)
```
ilumsden: Before calling `_fill_perfdata`, I would filter the dataframe down to just the columns that will be involved in the printing. Otherwise, you could be filling in the entire dataframe when you only need to fill in one or two columns.

thicket/utils.py
```diff
@@ -256,3 +258,36 @@ def validate_nodes(tk):
...
+    Arguments:
+        df (DataFrame): DataFrame to fill missing rows in
+        numerical_fill_value (any): value to fill numerical rows with
+
+    Returns:
+        (DataFrame): filled DataFrame
+    """
+    try:
+        # Fill missing rows in dataframe with NaN's
+        df = df.reindex(
```
ilumsden: Do not reassign to `df` here. By doing that, all changes you make in this function will be propagated back to the input. For example, when you call `_fill_perfdata(self.dataframe)` in `Thicket.tree`, this line of code causes you to fill in the dataframe in the Thicket object, not just a copy.

ilumsden (changes requested):
Mostly looks good. Just a couple of minor changes to ensure good performance and prevent unintended modification of the Thicket data structure.

thicket/utils.py
michaelmckinsey1: [83ac9cf](https://github.com/LLNL/thicket/pull/181/commits/83ac9cf2c6381bc1ae4a88e704c7599866622e05) fixes this

thicket/thicket.py
michaelmckinsey1: tldr: checking for these columns is not pretty, but there is up to 20% performance improvement

I tried this in 650007a. It is not trivial to select the columns, because there are 4 different dataframe columns as arguments to tree, default arguments may not exist, and `metric_column` can be a `list` or a `str`.

I collected some performance numbers below for this specific change. I tried to make up examples where `_fill_perfdata` would be doing significant work.

### Testing how long Thicket.tree takes to return (without printing)

#### 1 LBANN file + adding arbitrary columns to stress test (6558 nodes, 110 cols)
without change (3 trials): 
- 3.34s, 3.36s, 3.45s

with 650007a (3 trials):
- 2.9s, 2.8s, 3.16s

#### Contrived example with two different datasets (1 LBANN + 16 AMG) + adding arbitrary columns (6575 nodes, 114 columns)
without: 3.25s
with 650007a: 2.56s

thicket/utils.py
```diff
@@ -256,3 +258,37 @@ def validate_nodes(tk):
...
+    """Create full index for DataFrame and fill created rows with NaN's or None's where applicable.
+
+    Arguments:
+        df (DataFrame): DataFrame to fill missing rows in
+        numerical_fill_value (any): value to fill numerical rows with
+
+    Returns:
+        (DataFrame): filled DataFrame
+    """
+    new_df = df.copy()
```
ilumsden: This is fine for now, although it may cause issues in the future if we ever start including data beyond POD. That's because `DataFrame.copy` with default arguments performs a shallow copy. So, any by-reference objects in `new_df` that are edited would be edited in `df` as well.

However, we currently don't deal with by-reference objects (besides the nodes, but that can be safely ignored in this case), and I can't see any scenario in the near future where we would. So, we should keep this in mind, but I don't think anything has to be done about this in this PR.

thicket/thicket.py
```diff
@@ -912,8 +913,34 @@ def tree(
...
+                df_cols = [
+                    col
+                    for col in [
+                        metric_column,
+                        annotation_column,
+                        name_column,
+                        context_column,
+                    ]
+                    if col in self.dataframe.columns
+                ]
```
ilumsden, line 950: If you really want to shorten the code (at the expense of some readability), you could do this:
```python
df_cols = [
    col
    for col in [
        *metric_column if isinstance(metric_column, list) else metric_column,
        annotation_column,
        name_column,
        context_column,
    ]
    if col in self.dataframe.columns
]
```

ilumsden (approved):
LGTM.

However, @pearce8, when you do your review, you should take a look at the comment @michaelmckinsey1 left regarding performance.

thicket/thicket.py
michaelmckinsey1: `*metric_column if isinstance(metric_column, list) else metric_column` isn't valid syntax here. closest I can get is:

```
tree_cols = (
    metric_column
    if isinstance(metric_column, list)
    else [metric_column]
    + [
        annotation_column,
        name_column,
        context_column,
    ]
)
df_cols = [col for col in tree_cols if col in self.dataframe.columns]
```

thicket/utils.py
michaelmckinsey1: Would the solution you propose be our own function for deepcopying Pandas DataFrames?

ilumsden: No, the solution would be `new_df = df.copy(deep=True)`. However, don't worry about adding that in this PR. We'll only run into this is we make pretty major changes to the types of data being stored in the dataframe. I can't see that happening in the foreseeable future.

thicket/thicket.py
ilumsden: Huh. You learn something new every day. Apparently, the `*` unpack operator has to apply to the entire expression. You could rewrite what I have above as the following, and it should work:
```python
df_cols = [
    col
    for col in [
        *(metric_column if isinstance(metric_column, list) else [metric_column]),
        annotation_column,
        name_column,
        context_column,
    ]
    if col in self.dataframe.columns
]
```

This works because the `*` operator is applied to all results of the ternary. To make sure that works correctly, I wrap the value in the else clause in square braces to force it to be a list that can be unpacked by `*`. Note that you **have** to use `[]` for this. If you use the `list()` constructor instead, the contents of your column name will be turned into a list. So, if the column name is `"time"`, using `[]` produces `["time"]`, but using `list()` produces `["t", "i", "m", "e"]`.

thicket/utils.py
michaelmckinsey1: I thought you were referring to the fact that `deep=True` [does not copy recursively](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.copy.html#:~:text=When%20deep%3DTrue%2C%20data%20is%20copied%20but%20actual%20Python%20objects%20will%20not%20be%20copied%20recursively%2C%20only%20the%20reference%20to%20the%20object.%20This%20is%20in%20contrast%20to%20copy.deepcopy%20in%20the%20Standard%20Library%2C%20which%20recursively%20copies%20object%20data%20(see%20examples%20below).). `deep=True` is the default so that change wouldn't do anything. 

However, `copy.deepcopy` is apparently recursive.

thicket/thicket.py
michaelmckinsey1: works for me [7d8b397](https://github.com/LLNL/thicket/pull/181/commits/7d8b397935395270ef96063d657e4adc6d720d78)

pearce8 (approved):

PR #182, opened 2024-06-24
Run Unit Tests for Different Parameters

michaelmckinsey1:
This PR proposes to expand running the unit tests for (1) Thickets with intersection trees (default is union) and (2) Thickets without filling the performance data (default is filling the performance data). Therefore, testing for each combination would run all the unit tests for 4 types of Thickets:

1. Union+Filled
2. Union+NonFilled
3. Intersection+Filled
4. Intersection+NonFilled

[Example of using parametrized fixtures from pytest docs](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#:~:text=%40pytest.fixture(params%3D%5B0%2C%201%5D%2C%20ids%3D%5B%22spam%22%2C%20%22ham%22%5D)%0Adef%20a(request)%3A%0A%20%20%20%20return%20request.param)

When tests fail, we can see which configuration of parameters failed:
```
FAILED thicket/tests/test_tree.py::test_indices[Intersection-FillPerfdata] - Key...
FAILED thicket/tests/test_tree.py::test_indices[Union-NoFillPerfdata] - KeyError: ...
```
We can run single tests for a single set of parameters by specifying them to pytest
```
$ python -m pytest thicket/tests/test_tree.py::test_indices[Union-NoFillPerfdata]

============================= test session starts =============================
platform win32 -- Python 3.11.7, pytest-8.0.0, pluggy-1.4.0
rootdir: C:\Users\Micha\Documents\Github\thicket
configfile: pytest.ini
collected 1 item

thicket\tests\test_tree.py F                                             [100%]

```


michaelmckinsey1:
Depends on ~#193~

michaelmckinsey1:
depends on ~#181~

thicket/tests/test_concat_thickets.py
```diff
@@ -90,38 +111,56 @@ def test_filter_concat_thickets_columns(thicket_axis_columns):
     filter_multiple_and(combined_th, columns_values)
 
 
-def test_filter_stats_concat_thickets_columns(thicket_axis_columns):
+def test_filter_stats_concat_thickets_columns(thicket_axis_columns, intersection):
```
ilumsden, line 114: Why is `fill_perfdata` not in this test case?

thicket/tests/test_concat_thickets.py
```diff
@@ -90,38 +111,56 @@ def test_filter_concat_thickets_columns(thicket_axis_columns):
...
+        new_col = less_than_20 + less_than_45 + less_than_75
+        combined_th.statsframe.dataframe[("test", "test_string_column")] = new_col
+        # set numeric column values
+        combined_th.statsframe.dataframe[("test", "test_numeric_column")] = range(0, 74)
 
     check_filter_stats(combined_th, columns_values)
 
 
-def test_query_concat_thickets_columns(thicket_axis_columns):
+def test_query_concat_thickets_columns(thicket_axis_columns, intersection):
```
ilumsden, line 141: Same here. Why is `fill_perfdata` not in this test case?

thicket/tests/test_intersection.py
```diff
@@ -7,11 +7,17 @@
 from thicket import Thicket as th
 
 
-def test_intersection(rajaperf_cali_1trial):
-    tk = th.from_caliperreader(rajaperf_cali_1trial, disable_tqdm=True)
-
+def test_intersection(rajaperf_cali_1trial, fill_perfdata):
```
ilumsden, line 10: I would ask why this test case doesn't have `intersection`, but this one makes sense

ilumsden (commented):
This looks good, but there are a couple of things I'd like you to clarify @michaelmckinsey1.

michaelmckinsey1:
@ilumsden Tests that use the `thicket_axis_columns` fixture are already parametrized, since `thicket_axis_columns` is parametrized itself. Unlike the other fixtures, which are lists of files, `thicket_axis_columns` and `stats_thicket_axis_columns` create the Thickets in the fixtures. 

So running `python -m pytest thicket/tests/test_concat_thickets.py::test_filter_stats_concat_thickets_columns` results in:
```
0.44s call     thicket/tests/test_concat_thickets.py::test_filter_stats_concat_thickets_columns[Intersection-FillPerfdata]
0.41s call     thicket/tests/test_concat_thickets.py::test_filter_stats_concat_thickets_columns[Intersection-NoFillPerfdata]
0.33s call     thicket/tests/test_concat_thickets.py::test_filter_stats_concat_thickets_columns[Union-FillPerfdata]
0.33s call     thicket/tests/test_concat_thickets.py::test_filter_stats_concat_thickets_columns[Union-NoFillPerfdata]
```

thicket/thicket.py
```diff
@@ -1147,15 +1147,33 @@ def intersection(self):
...
+                keep = all([df[header].notna() # We are checking for NaNs
+                            .all() # For all values in a row
+                            .all() # For all rows in the slice
+                            for header in self.dataframe.columns.get_level_values(0)] # For each column header df[header] == slice
+                        )
+                rows.extend([keep]*extend_len) # Extend by extend_len for MultiIndex
+            tkc = self.deepcopy()
+            tkc.dataframe = tkc.dataframe[rows]
+            tkc = tkc.squash()
+            return tkc
```
michaelmckinsey1: @ilumsden Do you think this can be a query? This code works but I was not successful at making a query.

This is performing an intersection when the columns are MultiIndex, which can be identified when all of the metric values for a given header are NA like the three rows seen below under the `l` header.

image

I don't recall if there is support for MultiIndex columns in the query language.

thicket/tests/test_concat_thickets.py
michaelmckinsey1: see https://github.com/LLNL/thicket/pull/182#issuecomment-2234018765

thicket/thicket.py
ilumsden: There is support for MultiIndex columns in the QL nowadays, but I would not recommend doing this with the QL unless it's really needed. If we care about performance, the QL shouldn't be used for everything. At the end of the day, the problem that the QL is solving (a special variant of subgraph isomorphism) is an NP-Hard problem. Although I can make the QL faster, it will never be super fast due to that fact.

The QL should be used when you need to select or filter a Thicket _while considering the node relationships encoded in the graph_. If you don't need to take those relationships into account, then the QL could introduce unnecessary bottlenecks or slowdowns into your code. A good example of when you _should_ use the QL is what you do in `NCUReader` @michaelmckinsey1. A good example of when you _should not_ use the QL is when you just want to apply a filter to each node independently.

Besides all that stuff related to the QL, I'd also recommend you don't do the mentioned change in this PR. That's a different change to the code, so it would belong in a different PR if you were to do it.

thicket/tests/conftest.py
```diff
@@ -36,6 +52,7 @@ def thicket_axis_columns(rajaperf_cali_1trial):
     combined_th = Thicket.concat_thickets(
         thickets=thickets,
         axis="columns",
+        calltree=["union", "intersection"][intersection],
```
ilumsden: Since we're not in C, please don't do the whole "True can be interpreted as 1 and False can be interpreted as 0" thing. This code does work, but it is unbelievably unintuitive. I had to stare at it for 5 minutes before I realized what you were doing.

The normal way to do this in Python is:
```python
calltree="intersection" if intersection else "union",
```

thicket/tests/conftest.py
```diff
@@ -71,6 +98,7 @@ def stats_thicket_axis_columns(rajaperf_cuda_block128_1M_cali):
...
```
ilumsden: See other comment about this syntax

thicket/thicket.py
```diff
@@ -1147,15 +1147,37 @@ def intersection(self):
...
-            query = Query().match(".", lambda df: len(df) == len(self.profile))
+            # Check for padded perfdata
+            if self.dataframe["name"].isnull().any():
+                # Row that didn't exist will contain "None" in the name column.
+                query = Query().match(
+                    ".", lambda df: df["name"].apply(lambda n: n is not None).all()
+                )
+            else:
+                # If perfdata not padded
+                query = Query().match(".", lambda df: len(df) == len(self.profile))
```
ilumsden, line 1220: I didn't see this before. I get why you were asking about using the QL with MultiIndex now. I'd still recommend not using the QL here. At the end of the day, it's almost always better to use _P_ code than _NP Hard_ code.

ilumsden (changes requested):
Got a few more changes that I'd like you to make. The most notable is that I'd encourage you to remove the use of the QL in `Thicket.intersection` since you are already editing it.

thicket/thicket.py
ilumsden: Ignore my point about not changing this in this PR. I hadn't gone over your changes before I said that. My comment about not using the QL willy-nilly stands though.

thicket/tests/conftest.py
michaelmckinsey1: sure 6df4876

thicket/tests/conftest.py
michaelmckinsey1: 6df4876

thicket/thicket.py
michaelmckinsey1: I suppose if there were a case where the parent was NaN, e.g.:

```
1.781 RAJAPerf
├─ nan Algorithm
│  ├─ 0.002 Algorithm_MEMCPY
│  ├─ 0.002 Algorithm_MEMSET
│  ├─ 0.003 Algorithm_REDUCE_SUM
```

becomes

```
1.781 RAJAPerf
├─ 0.002 Algorithm_MEMCPY
├─ 0.002 Algorithm_MEMSET
├─ 0.003 Algorithm_REDUCE_SUM
```

when it should be 

```
1.781 RAJAPerf
```

I can't think of a situation where this would happen, since in reality if the parent node was filled with NaNs because it did not exist, then the children shouldn't have existed either. The children can't exist without the parent, so they would also have NaNs. However, the QL would properly remove the children in this example.

michaelmckinsey1: I prefer using the QL where I can, because it is better tested and more readable than writing new code like in this PR. I suppose then you would think that we would benefit from having a performance data filter function, and not from the QL? And a performance data filter would replace the code in `intersection`, including the existing queries?

ilumsden: Going back and trying to knock out some of these reviews now.

A performance data filter function (similar to Hatchet's `filter` function) would be a great thing to have. At the end of the day, if you are doing filtering that doesn't care about the structure of the graph, you really should be using Pandas operations over the QL, and that's what this type of function would provide. I have some thoughts on how to improve the QL's performance, but it should still never be faster than Pandas on single node filtering (unless Pandas is **much** worse in performance than I think).

That being said I do get your point about using the QL in this case. I think a good thing to do here would be to continue using the QL in `intersection` (for now at least, we may revisit later) and then create a performance data filter function.

ilumsden (approved):
LGTM, but see my last comment regarding performance data filtering and the QL.

slabasan:
@michaelmckinsey1 Can you please resolve conflicts on this PR?

slabasan (approved):

PR #185, opened 2024-06-26
Optimize Loop in Unify

michaelmckinsey1:
This PR:

1. Refactors `_merge_dicts` and `_replace_graph_df_nodes` by removing the functions and putting the code inline.
2. Optimizes the double for loop that merges `old_to_new` and `temp_dict` from `O(n^2)` to `O(n)`. This was an oversight during original implementation.

The performance improvements of this PR are documented in #170. #170 avoids needing to merge dictionaries as it only calls `concat_thickets` with 2 Thickets, however using the `concat_thickets` function with 3 or more Thickets will still benefit from this PR. 

This PR is better for fewer files with many nodes, while #170 is better for many files since it avoids merging `old_to_new` and `temp_dict`.

### DIFF
--- a/thicket/ensemble.py
+++ b/thicket/ensemble.py
@@ -54,12 +54,12 @@ class Ensemble:
             # This is necessary to avoid applying updates to the DataFrames every iteration.
             # Merge values of new_dict into keys of old_to_new
             merged_dict = {}
-            seen_keys = []
+            seen_keys = set()
             for old_id, cur_node in old_to_new.items():
                 cur_id = id(cur_node)
                 if cur_id in new_dict:
                     merged_dict[old_id] = new_dict[cur_id]
-                    seen_keys.append(cur_id)
+                    seen_keys.add(cur_id)            
             # Add pairs that are left from new_dict into old_to_new
             for cur_id, new_node in new_dict.items():
                 if cur_id not in seen_keys:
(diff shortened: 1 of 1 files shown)


thicket/ensemble.py
```diff
@@ -111,13 +53,35 @@ def _replace_graph_df_nodes(thickets, old_to_new, union_graph):
...
-            # Merge the current old_to_new dictionary with the new mappings
-            old_to_new = _merge_dicts(old_to_new, temp_dict)
+            # Merge the current old_to_new dictionary with the new mappings.
+            # This is necessary to avoid applying updates to the DataFrames every iteration.
+            # Merge values of temp_dict into keys of old_to_new
+            seen_keys = []
+            for cur_id, cur_node in old_to_new.items():
+                new_id = id(cur_node)
+                if new_id in temp_dict:
+                    old_to_new[cur_id] = temp_dict[new_id]
```
ilumsden: Never update the thing you are iterating over. You may get lucky, but 9 times out of 10, this is at best undefined behavior.

ilumsden (changes requested):
Mostly looks good. Just one issue that needs to be addressed to ensure we don't run into unexpected (and possibly undefined) behavior.

thicket/ensemble.py
michaelmckinsey1: Good catch I messed up when converting it from a function [a1479b3](https://github.com/LLNL/thicket/pull/185/commits/a1479b3094970375e92c378f95dc169207610518)

ilumsden (approved):
LGTM

ilumsden:

pearce8 (approved):

PR #190, opened 2024-07-05
Fix Deepcopy Bug

michaelmckinsey1:
In Thicket, we assume 
```
self.graph is self.statsframe.graph
```
current `deepcopy()` breaks this assumption and this is a fix.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -846,7 +846,7 @@ class Thicket(GraphFrame):
             profile=copy.deepcopy(self.profile),
             profile_mapping=copy.deepcopy(self.profile_mapping),
             statsframe=GraphFrame(
-                graph=gf.graph, dataframe=self.statsframe.dataframe.copy()
+                graph=gf.graph, dataframe=self.statsframe.dataframe.copy(deep=True)
             ),
             statsframe_ops_cache=self.statsframe_ops_cache.copy(),
         )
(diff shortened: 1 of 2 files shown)


thicket/thicket.py
```diff
@@ -845,7 +845,9 @@ def deepcopy(self):
             metadata=self.metadata.copy(),
             profile=copy.deepcopy(self.profile),
             profile_mapping=copy.deepcopy(self.profile_mapping),
-            statsframe=self.statsframe.deepcopy(),
+            statsframe=GraphFrame(
+                graph=gf.graph, dataframe=self.statsframe.dataframe.copy()
```
ilumsden: Since this is a deepcopy operation, we should use a dataframe deepcopy too. Change the dataframe piece of this line to:
```python
dataframe=self.statsframe.dataframe.copy(deep=True)
```

ilumsden (changes requested):
Good catch. I've got one very minor change that should be made, but looks good otherwise.

thicket/thicket.py
michaelmckinsey1: It is [deep by default](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.copy.html). ~Is this just to be explicit?~ We are going to be explicit.

ilumsden (approved):
LGTM

pearce8 (approved):

PR #192, opened 2024-07-09
Add Root Node to Graph

michaelmckinsey1:
Examples of `add_root_node` and `get_node` can be found in the notebook in [thicket-tutorial/48](https://github.com/LLNL/thicket-tutorial/blob/parallel-sorting/notebooks/08A_composing_parallel_sorting_data.ipynb)

`add_root_node` enables adding a root node to the `Thicket.graph` and corresponding `Thicket.statsframe` and `Thicket.dataframe` objects.

`get_node` provides a simpler interface for grabbing a node object by its name in the Thicket instead of using something like `[n for n in self.graph.traverse() if n.frame["name"] == name][0]`. This function is optional for this PR or could be its own PR; I am proposing it here.

### DIFF
--- a/thicket/tests/test_add_root_node.py
+++ b/thicket/tests/test_add_root_node.py
@@ -1,0 +1,26 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+from hatchet.node import Node
+
+
+def test_add_root_node(literal_thickets):
+    tk, tk2, tk3 = literal_thickets
+
+    assert len(tk.graph) == 4
+
+    tk.add_root_node({"name": "Test", "type": "function"})
+
+    test_node = tk.get_node("Test")
+
+    # Check if node was inserted in all components
+    assert isinstance(test_node, Node)
+    assert len(tk.graph) == 5
+    assert len(tk.statsframe.graph) == 5
+    assert test_node in tk.dataframe.index.get_level_values("node")
+    assert test_node in tk.statsframe.dataframe.index.get_level_values("node")
+
+    assert tk.dataframe.loc[test_node, "name"].values[0] == "Test"
+    assert tk.statsframe.dataframe.loc[test_node, "name"] == "Test"
--- a/thicket/tests/test_add_root_node.py
+++ b/thicket/tests/test_add_root_node.py
@@ -7,7 +7,7 @@ from hatchet.node import Node
 
 
 def test_add_root_node(literal_thickets):
-    tk, tk2, tk3 = literal_thickets
+    tk, _, _ = literal_thickets
 
     assert len(tk.graph) == 4
 
--- a/thicket/tests/test_add_root_node.py
+++ b/thicket/tests/test_add_root_node.py
@@ -11,12 +11,15 @@ def test_add_root_node(literal_thickets):
 
     assert len(tk.graph) == 4
 
+    # Call add_root_node
     tk.add_root_node({"name": "Test", "type": "function"})
-
+    # Get node variable
     test_node = tk.get_node("Test")
 
     # Check if node was inserted in all components
     assert isinstance(test_node, Node)
+    assert test_node._hatchet_nid == 3
+    assert test_node._depth == 0
     assert len(tk.graph) == 5
     assert len(tk.statsframe.graph) == 5
     assert test_node in tk.dataframe.index.get_level_values("node")
--- a/thicket/tests/test_get_node.py
+++ b/thicket/tests/test_get_node.py
@@ -9,7 +9,7 @@ import pytest
 def test_get_node(literal_thickets):
     tk, _, _ = literal_thickets
 
-    with pytest.raises(ValueError):
+    with pytest.raises(KeyError):
         tk.get_node("Foo")
 
     baz = tk.get_node("Baz")
--- a/thicket/tests/test_get_node.py
+++ b/thicket/tests/test_get_node.py
@@ -7,7 +7,7 @@ import pytest
 
 
 def test_get_node(literal_thickets):
-    tk, tk2, tk3 = literal_thickets
+    tk, _, _ = literal_thickets
 
     with pytest.raises(ValueError):
         tk.get_node("Foo")
--- a/thicket/tests/test_get_node.py
+++ b/thicket/tests/test_get_node.py
@@ -10,7 +10,7 @@ def test_get_node(literal_thickets):
     tk, tk2, tk3 = literal_thickets
 
     with pytest.raises(ValueError):
-        foo = tk.get_node("Foo")
+        tk.get_node("Foo")
 
     baz = tk.get_node("Baz")
 
--- a/thicket/tests/test_get_node.py
+++ b/thicket/tests/test_get_node.py
@@ -1,0 +1,20 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+import pytest
+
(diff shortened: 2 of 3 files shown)


michaelmckinsey1:
Depends on ~#190 and #173~

thicket/thicket.py
```diff
@@ -1508,6 +1512,53 @@ def get_unique_metadata(self):
...
+        """Add node at root level with given attributes.
+
+        Arguments:
+            attrs (dict): attributes for the new node which will be used to initilize the
+            node.frame.
+        """
+
+        new_node = hatchet.node.Node(
+            frame_obj=frame.Frame(attrs=attrs), hnid=len(self.graph)
+        )
```
ilumsden: If you're going to do `frame.Frame`, why not also do `node.Node`? Or, better yet, just do the imports in a way that let's you do `Node` and `Frame`.

thicket/thicket.py
```diff
@@ -1508,6 +1512,53 @@ def get_unique_metadata(self):
...
+        new_node_df = pd.DataFrame(
+            index=pd.MultiIndex.from_product(new_idx, names=idx_levels)
+        )
+        new_node_df["name"] = attrs["name"]
+        self.dataframe = pd.concat([self.dataframe, new_node_df])
+
+        # statsframe.dataframe
+        self.statsframe.dataframe = helpers._new_statsframe_df(self.dataframe)
+        # Reapply stats operations after clearing statsframe dataframe
+        self.reapply_stats_operations()
```
ilumsden, line 1547: I have one issue and one thing I think is odd with this function.

First, this function will not properly set the `_depth` field of the `Node`. This won't cause issues with much of Thicket, but there are a few places (e.g., certain queries with query language) where this could cause issues.

Second, it's odd that this function makes the new root node have the largest `_hatchet_nid` field in the `Thicket` object. It's not an issue, but it is odd.

There are two functions from Hatchet that can help with this. `hatchet.Graph.enumerate_depth` will traverse the graph and update the `_depth` field for all nodes. `hatchet.Graph.enumerate_traverse` will traverse the graph and update the `_hatchet_nid` fields for all nodes. `enumerate_traverse` also calls `enumerate_depth`.

thicket/thicket.py
```diff
@@ -1508,6 +1512,53 @@ def get_unique_metadata(self):
...
+        Returns:
+            (node): Hatchet Node object
+        """
+        node = [n for n in self.graph.traverse() if n.frame["name"] == name]
+
+        if len(node) > 1:
+            warnings.warn(f'More than one node with name "{name}". Returning a list')
+            return node
+
+        return node.pop()
```
ilumsden: It would be better to just return `node[0]`. `pop` will do things like resizing (see the [source code](https://github.com/python/cpython/blob/5c22476c01622f11b7745ee693f8b296a9d6a761/Objects/listobject.c#L1021)), which isn't needed.

ilumsden (changes requested):
Mostly looks good.

The only change that really _needs_ to be made is the one about `pop()`. The other two are just suggestions.

thicket/thicket.py
pearce8: I am not worried about the largest nid, but the depth should be verified and correct for all of the nodes after this operation.

More importantly, I am not sure about the interface.  If adding a root node, what becomes its children?  All the nodes that were previously root nodes?

thicket/thicket.py
michaelmckinsey1: [e2d957c](https://github.com/LLNL/thicket/pull/192/commits/e2d957cd265d0b6b25bbba6c7d949a0fcbf761ed)

thicket/thicket.py
michaelmckinsey1: I was using `node.Node` but we use `node` as a variable name in squash and I'm using it in `get_node()`. Not a big deal, so I'll just import `Node` and `Frame`

[7e56d11](https://github.com/LLNL/thicket/pull/192/commits/7e56d11403fc4f7ca4cf848cf0ae19076700d538)

thicket/thicket.py
michaelmckinsey1: Good catch, depth was not being set. Now calling `self.graph.enumerate_traverse()` to set depth and hatchet nid.

michaelmckinsey1: > I am not worried about the largest nid, but the depth should be verified and correct for all of the nodes after this operation.
> 
> More importantly, I am not sure about the interface. If adding a root node, what becomes its children? All the nodes that were previously root nodes?

Children would not be a concern in this case, since we are adding a root node with no children and no parents. The other root nodes will be the siblings of the new root node. Adding A node anywhere in the tree would be more involved than this function.

thicket/tests/test_add_root_node.py
```diff
@@ -0,0 +1,29 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+from hatchet.node import Node
+
+
+def test_add_root_node(literal_thickets):
+    tk, tk2, tk3 = literal_thickets
```
ilumsden: Since you're not using `tk2` or `tk3`, change this line to the following:

```python
tk, _, _ = literal_thickets
```

thicket/tests/test_get_node.py
```diff
@@ -0,0 +1,20 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+import pytest
+
+
+def test_get_node(literal_thickets):
+    tk, tk2, tk3 = literal_thickets
```
ilumsden: See previous comment about use of underscores here

thicket/thicket.py
```diff
@@ -1514,6 +1516,59 @@ def get_unique_metadata(self):
...
+        Returns:
+            (node): Hatchet Node object
+        """
+        node = [n for n in self.graph.traverse() if n.frame["name"] == name]
+
+        if len(node) > 1:
+            warnings.warn(f'More than one node with name "{name}". Returning a list')
+            return node
+        elif len(node) == 0:
+            raise ValueError(f'Node with name "{name}" not found.')
```
ilumsden: I would raise a `KeyError` here instead.

thicket/thicket.py
```diff
@@ -1514,6 +1516,59 @@ def get_unique_metadata(self):
...
+        Arguments:
+            name (str): name of the node (Node.frame['name']).
+
+        Returns:
+            (node): Hatchet Node object
+        """
+        node = [n for n in self.graph.traverse() if n.frame["name"] == name]
+
+        if len(node) > 1:
+            warnings.warn(f'More than one node with name "{name}". Returning a list')
```
ilumsden: I don't think this warning is necessary as long as we document the return type correctly. Regardless if this warning is kept or removed, the docstring needs to be updated to indicate that it can return either `Node` or `List[Node]`.

ilumsden (changes requested):
This mostly looks good at this point. Just a couple of minor changes.

thicket/tests/test_add_root_node.py
michaelmckinsey1: 53f8718

thicket/tests/test_get_node.py
michaelmckinsey1: 53f8718

thicket/thicket.py
michaelmckinsey1: 5053f83

thicket/thicket.py
michaelmckinsey1: fcc266a

ilumsden (approved):
LGTM.

pearce8 (approved):

Issue #186, opened 2024-06-26
Intersection Broken When _fill_perfdata=False


michaelmckinsey1:
Found in #182. Current `intersection` relies on the `None`s inserted by `_fill_perfdata` to compute the intersection after the union has already been applied

michaelmckinsey1:
Fixed in #193

PR #193, opened 2024-07-09
Fix intersection

michaelmckinsey1:
### Fix the intersection function when Thicket was created using `_fill_perfdata=False`.

Add separate query for when the perfdata **is not padded** as checking the name column for `None` values will not work. Additionally, the new query that checks the length of per-node slices of the perfdata will not work when the perfdata **is padded**, so both queries are necessary.


michaelmckinsey1:
Fixes #186

ilumsden (approved):
LGTM.

pearce8 (approved):

PR #194, opened 2024-07-10
Remove Optional Argument from _fill_perfdata

michaelmckinsey1:
Always fill numerical columns with `np.nan` to make functionality clearer.

### DIFF
--- a/thicket/utils.py
+++ b/thicket/utils.py
@@ -261,7 +261,7 @@ def validate_nodes(tk):
 
 
 def _fill_perfdata(df):
-    """Create full index for DataFrame and fill created rows with NaN's or None's where applicable.
+    """Create full index for DataFrame and fill created rows with NaN's or Nones where applicable.
 
     Arguments:
         df (DataFrame): DataFrame to fill missing rows in
--- a/thicket/utils.py
+++ b/thicket/utils.py
@@ -260,27 +260,28 @@ def validate_nodes(tk):
         )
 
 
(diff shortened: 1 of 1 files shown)


ilumsden (approved):
Looks good.

One minor comment is that it's probably better to use `NaN` in the comments, but that's so minor that it's fine if you don't do that.

Also, whenever it's no longer experimental in pandas, we should probably switch to using `pandas.NA`. It's a pandas-specific type for representing missing data that works for all pandas data types. So, `NA` can be interpreted as `Int64Dtype`, `StringDtype`, `Float64Dtype`, `BooleanDtype`, and even the newer `ArrowDtype`.

pearce8 (approved):

michaelmckinsey1:
> Looks good.
> 
> One minor comment is that it's probably better to use `NaN` in the comments, but that's so minor that it's fine if you don't do that.
> 
> Also, whenever it's no longer experimental in pandas, we should probably switch to using `pandas.NA`. It's a pandas-specific type for representing missing data that works for all pandas data types. So, `NA` can be interpreted as `Int64Dtype`, `StringDtype`, `Float64Dtype`, `BooleanDtype`, and even the newer `ArrowDtype`.
> 
> @pearce8 this is ready for your review

I reverted the change back to "NaN". They actually were "NaN" before and I had changed it to "nan" to match `np.nan`. However, since you mentioned it this doesn't make sense, as we should only be using `NaN` or `np.nan`, and `NaN` makes more sense in the context of comments.

pearce8 (approved):

PR #196, opened 2024-07-17
Fix Bug in add_ncu

michaelmckinsey1:
#178 Introduced a bug when `chosen_metrics` was not provided. When `chosen_metrics=None` all ncu metrics are added to the perf dataframe. This PR fixes the bug by modifying the check and moving it, where we can check for duplicate columns when `chosen_metrics=None` aswell.

- Check type of `chosen_metrics` before reading NCU
- Check for duplicate columns after reading NCU, such that if `chosen_metrics` is `None` we will know which metrics are present.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -590,15 +590,6 @@ class Thicket(GraphFrame):
             warnings.warn(f"Removing duplicate metrics in chosen_metrics: {dupe_mets}")
         chosen_metrics = unique_metrics
 
-        # Check if chosen_metrics are in the dataframe
-        dupe_cols = [col for col in chosen_metrics if col in self.dataframe.columns]
-        if overwrite:
-            self.dataframe = self.dataframe.drop(columns=dupe_cols)
-        elif not overwrite and len(dupe_cols) > 0:
-            raise ValueError(
-                f"Columns {dupe_cols} already exist in the performance data table. Set overwrite=True to overwrite."
-            )
-
         # Initialize reader
         ncureader = NCUReader()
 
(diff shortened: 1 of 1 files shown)


thicket/thicket.py
```diff
@@ -602,6 +593,15 @@ def _rep_agg_func(col):
         if chosen_metrics:
             ncu_df = ncu_df[chosen_metrics]
```
ilumsden, line 624: You need to have a type check before doing this. Since you don't validate the type of `chosen_metrics` anywhere, you could get an input that would cause various parts of the rest of the code to error. Off the top of my head, some scenarios are:
* Passing a number (e.g., `int`, `float`) would cause this line to fail with a `KeyError` iff there is not a column with an integer or float index
* Passing a string would cause this line to succeed, but it would also return a `Series` instead of a `DataFrame`. This would cause all the code following this line to fail.

ilumsden (changes requested):
The revised logic for `dupe_cols` looks good, but there's a pretty major issue in handling `chosen_metrics` that could easily break `NCUReader` if not addressed.

thicket/thicket.py
michaelmckinsey1: [b7cc225](https://github.com/LLNL/thicket/pull/196/commits/b7cc2252b7ce94527c483ac59697607d7ec713e1)

ilumsden (approved):
LGTM.

slabasan:
@michaelmckinsey1 Can you please resolve conflicts on this PR? Apologies if I merged these in the wrong order.

slabasan (approved):

PR #197, opened 2024-07-22
Fix Indexing Issue in Tree for MultiIndex

michaelmckinsey1:
#181 introduced a bug for MultiIndex columns on this [line](https://github.com/LLNL/thicket/blob/72fca533bc1d3980c9e7865126aeb826bdc58a47/thicket/thicket.py#L955), when trying to index the dataframe. The user can specifying the name column using a tuple or we automatically select the right column when `name_column` is the default argument. This PR also adds a unit test for this case so we can tell if another change breaks MultiIndex printing.

### DIFF
--- a/thicket/external/console.py
+++ b/thicket/external/console.py
@@ -258,10 +258,7 @@ class ThicketRenderer(ConsoleRenderer):
                 else:
                     metric_str += " [{}]".format(annotation_content)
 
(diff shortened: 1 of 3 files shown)


ilumsden (approved):
LGTM.

@pearce8 please review and merge as soon as you are able. I am actively running into the bug that this fixes in the DDR vs HBM work.

pearce8 (approved):

PR #201, opened 2024-07-29
NCU Reader Support for RAJA_CUDA and Lambda_CUDA

michaelmckinsey1:
tldr:
- Support `Lambda_CUDA` and `RAJA_CUDA` variants by using demangled kernel names.
- Add `debug` flag to see detailed information about kernel matches, so we can preliminarily investigate future issues without editing the source code.
- Support cub kernels via string similarity to compare function signatures.
- Adds unit testing by refactoring the matching functions

### Description

Enables support for reading NCU report profiles for `RAJA_CUDA` and `Lambda_CUDA` variants and `cub` kernels by using the demangled action name.

The current Thicket NCU reader matches nodes in a Caliper `cuda_activity_profile` (CAP) and NCU report file by checking if an action in the report has the name `action.name(_ncu_report.IAction_NameBase_FUNCTION)`, which for `Base_CUDA` is the name of the kernel (e.g. `daxpy`, `energy1`, or `energy2`). This name can be found in the CAP node name `kernel_name in node.frame["name"]`.

For `RAJA_CUDA` and `Lambda_CUDA`, the above assumption does not hold, as the values for `action.name(_ncu_report.IAction_NameBase_FUNCTION)` will not be the kernel names. However, the kernel names are still embedded in the `action.name(_ncu_report.IAction_NameBase_DEMANGLED)` demangled action name. This PR parses the demangled name to match the nodes in the CAP, which also works for `Base_CUDA` profiles.

For cub kernels, there may be kernels with the same name, but different function signatures. For example, matching the ncu kernel `void cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<double, double, int>::Policy700, false, false, double, double, int>(const T4 *, T4 *, const T5 *, T5 *, T6 *, T6, int, int, cub::GridEvenShare<T6>)` to the first `DeviceRadixSortDownsweepKernel` in the following calltree:
```
nan RAJAPerf
└─ nan Algorithm
   ├─ nan Algorithm_SORT
   │  ├─ nan cudaLaunchKernel
   │  │  ├─ 1016096.000 void cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<double, cub::NullType, int>::Policy700, **false, false**, double, cub::NullType, int>(double const*, double*, cub::NullType const*, cub::NullType*, int*, int, int, int, cub::GridEvenShare<int>)
   │  │  ├─ 1399520.000 void cub::DeviceRadixSortDownsweepKernel<cub::DeviceRadixSortPolicy<double, cub::NullType, int>::Policy700, **true, false**, double, cub::NullType, int>(double const*, double*, cub::NullType const*, cub::NullType*, int*, int, int, int, cub::GridEvenShare<int>)
```
We use similarity matching using the standard library difflib `SequenceMatcher` to match the two, after first narrowing the search down to the `Algorithm_SORT` part of the calltree.

#### This PR (#201)
|  | Base_CUDA | Lambda_CUDA | RAJA_CUDA |
| -------- | ------- | ------- | ------- |
| rajaperf kernels | ✓ | ✓ | ✓ |
| cub kernels | ✓| ✓| ✓|
| kernels with multiple instances | ✓ | ✓ | ✓ |

#### Develop
|  | Base_CUDA | Lambda_CUDA | RAJA_CUDA |
| -------- | ------- | ------- | ------- |
| rajaperf kernels | ✓ | x | x |
| cub kernels | x| x | x|
| kernels with multiple instances | ✓ | x | x |


ilumsden:
Changing this PR to "work in progress" because there are complications due to mismatches in demangled kernel names from Caliper and NCU.

michaelmckinsey1:
This PR supersedes https://github.com/LLNL/thicket/tree/rajaperf-paper branch used for the rajaperf paper.

michaelmckinsey1:
> Changing this PR to "work in progress" because there are complications due to mismatches in demangled kernel names from Caliper and NCU.

Addressed in new changes.

thicket/ncu.py
```diff
@@ -124,35 +145,124 @@ def _read_ncu(thicket, ncu_report_mapping):
                     if len(kernel_call_trace) == 0:
                         continue
                     else:
+                        call_trace_found = True
+
+                        # Call trace with last element removed
+                        # (last elem usually not useful for matching)
+                        temp_call_trace = kernel_call_trace[:-1]
+                        # Special case to match "cub" kernels
+                        if "cub" in demangled_kernel_name:
```
dyokelson: Didn't we decide to try to remove the direct string matching and change to if multiple kernels are returned?

thicket/ncu.py
```diff
@@ -124,35 +145,124 @@ def _read_ncu(thicket, ncu_report_mapping):
...
+                        else:
+                            call_trace_str = "::".join(
+                                [s.lower() for s in temp_call_trace]
+                            )
+                        if debug:
+                            print(f"\tKernel Call Trace: {kernel_call_trace}")
+                            print(action.name())
+
+                        # Pattern ends with ":" if RAJA_CUDA, "<" if Base_CUDA
+                        kernel_pattern = rf"{call_trace_str}::(\w+)[<:]"
```
dyokelson: This feels fragile...and that things could change on us and break this, maybe we can think through other options for matching here.

dyokelson (changes requested):
Please add a pytest test for each of the cases in your table so we can make sure each is supported (Base_CUDA, Lambda_CUDA, RAJA_CUDA X rajaperf, cub, multiple

thicket/ncu.py
michaelmckinsey1: Even for single kernels, the cub kernels have `cub` in the name instead of the call trace. If we wait until there are multiple kernels returned, we would still have to check for `cub` with this logic.

thicket/ncu.py
michaelmckinsey1: The regex is only used to match a node or nodes in a subset of the tree, so typically a few nodes. They are not in the same order in the NCU report and in the call tree so we can not match based on ordering. Even more ambitious is the [sequence matching](https://github.com/michaelmckinsey1/thicket/blob/ff13797d08952f971dddc299ac6ca8c660cb988a/thicket/ncu.py#L220-L232) for the function signatures of the cub kernels, because they may have the same function names. Since all we have to use to match are the call traces (which are solid using query) and the demangled names (regex part), we have to parse the names in many cases.

michaelmckinsey1: Unit testing added in most recent commits [9a9eab0](https://github.com/LLNL/thicket/pull/201/commits/9a9eab01cfc5b7487e1cebb09906eef02e61308f)

slabasan:
> This PR supersedes https://github.com/LLNL/thicket/tree/rajaperf-paper branch used for the rajaperf paper.

Do you want to remove the `rajaperf-paper` branch on this repo?

thicket/ncu.py
slabasan: @dyokelson Can you take one last look at the unit tests to make sure your comments are addressed?

michaelmckinsey1:
> > This PR supersedes https://github.com/LLNL/thicket/tree/rajaperf-paper branch used for the rajaperf paper.
> 
> Do you want to remove the `rajaperf-paper` branch on this repo?

Done

thicket/ncu.py
```diff
@@ -113,8 +237,12 @@ def _read_ncu(thicket, ncu_report_mapping):
                 pbar = tqdm(range)
```
dyokelson: Can they disable tqdm?

dyokelson (commented):
looks good, just the question about disabling tqdm, otherwise will approve

thicket/ncu.py
dyokelson: lgtm

thicket/ncu.py
michaelmckinsey1: Option added in [0ef7c0d](https://github.com/LLNL/thicket/pull/201/commits/0ef7c0d8cfe78abd78230f5569089eecea4ed220)

dyokelson (approved):
LGTM

slabasan (approved):

PR #202, opened 2024-08-02
Check And Remove Duplicates in chosen_metrics

michaelmckinsey1:
@Yejashi Found a case where trying to add the same metric twice causes a bug.

Ex:
```
ncu_metrics = [
    "gpu__time_duration.sum",
    "gpu__time_duration.sum",
    "sm__throughput.avg.pct_of_peak_sustained_elapsed",
    "smsp__maximum_warps_avg_per_active_cycle",
    "smsp__maximum_warps_avg_per_active_cycle",
    "smsp__maximum_warps_avg_per_active_cycle",
]
```

is correctly handled now and prints a warning:
```
UserWarning: Removing duplicate metrics in chosen_metrics: ['gpu__time_duration.sum', 'smsp__maximum_warps_avg_per_active_cycle']
```

This is different from the functionality that checks if metrics in `chosen_metrics` are already in the performance data.

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: MIT
 
+import collections
 import copy
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


thicket/thicket.py
```diff
@@ -572,6 +573,17 @@ def _rep_agg_func(col):
...
                 return col[0]
 
+        # Remove duplicate metrics in chosen_metrics if the user provided duplicates
+        unique_metrics = list(set(chosen_metrics))
+        if len(unique_metrics) != len(chosen_metrics):
+            dupe_mets = [
+                met
+                for met, count in collections.Counter(chosen_metrics).items()
+                if count > 1
+            ]
```
ilumsden, line 583: This works, although it's not very efficient (`collections.Counter` is known for being pretty slow). That being said, we have other inefficiencies that we'll run into before this becomes a meaningful issue.

ilumsden (approved):
LGTM.

I have one comment about the performance of identifying the duplicates, but, IMO, it's not a big enough thing to block up this PR.

slabasan (approved):

PR #203, opened 2024-08-07
Fix Bug When Reading Directories

michaelmckinsey1:
#163 Introduced a bug when providing a directory of Caliper files that causes an error. Providing a list of files or a single file still works. This PR adds regression tests for this so we will catch it next time.


ilumsden (approved):
Good catch. LGTM.

PR #207, opened 2024-08-26
Statsframe Cache Pickle Serialization Fix

Yejashi:
This pr resolves a serialization issue caused by the `statsframe_ops_cache`.

The error is presented as follows:
```python
PicklingError: Can't pickle <function mean at 0x7fff9e7c7920>: it's not the same object as thicket.stats.mean.mean
```

The current fix it to reset the cache at every `to_pickle` call.


ilumsden (approved):
LGTM

pearce8 (approved):

PR #213, opened 2024-10-01
Confidence Interval Stats Function

### LINKED ISSUE #210: Add a function for calculatung confidence interval

Yejashi:
This PR introduces a new functionality to calculate confidence intervals within a Thicket object. The added confidence_interval function enables users to assess the variability and uncertainty in the performance data across multiple runs, providing a statistical range within which the true performance metric is likely to fall with a specified level of confidence.

### DIFF
--- a/thicket/stats/confidence_interval.py
+++ b/thicket/stats/confidence_interval.py
@@ -29,7 +29,7 @@ def confidence_interval(thicket, columns=None, confidence_level=0.95):
         columns (list): List of hardware/timing metrics to perform confidence interval
             calculation on. Note, if using a columnar_joined thicket a list of tuples
             must be passed in with the format (column index, column name).
-        confidence_level (int,float):  The confidence level (often 0.90, 0.95, or 0.99)
+        confidence_level (float):  The confidence level (often 0.90, 0.95, or 0.99)
             indicates the degree of confidence that the true parameter lies within the interval.
 
     Returns:
@@ -44,9 +44,7 @@ def confidence_interval(thicket, columns=None, confidence_level=0.95):
         raise ValueError("Value passed to 'columns' must be of type list.")
 
     if not isinstance(confidence_level, float):
-        raise ValueError(
-            r"Value passed to 'confidence_level' must be of type float or int."
-        )
+        raise ValueError(r"Value passed to 'confidence_level' must be of type float.")
 
     if confidence_level >= 1 or confidence_level <= 0:
         raise ValueError(
--- a/thicket/stats/confidence_interval.py
+++ b/thicket/stats/confidence_interval.py
@@ -1,0 +1,98 @@
+# Copyright 2022 Lawrence Livermore National Security, LLC and other
+# Thicket Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: MIT
+
+import numpy as np
+import pandas as pd
+import scipy.stats
+
+import thicket as th
+from ..utils import verify_thicket_structures
+from .stats_utils import cache_stats_op
+
+
+@cache_stats_op
+def confidence_interval(thicket, columns=None, confidence_level=0.95):
+    r"""Calculate the confidence interval for each node in the performance data table.
+
+    Designed to take in a thicket, and append one or more columns to the aggregated
+    statistics table for the confidence interval calculation for each node.
+
+    A confidence interval is a range of values, derived from sample data, that is
+    likely to contain the true population parameter with a specified level of confidence.
+    It provides an estimate of uncertainty around a sample statistic, indicating how much
+    variability is expected if the sampling process were repeated multiple times.
+
+    Arguments:
+        thicket (thicket): Thicket object
+        columns (list): List of hardware/timing metrics to perform confidence interval
+            calculation on. Note, if using a columnar_joined thicket a list of tuples
+            must be passed in with the format (column index, column name).
+        confidence_level (int,float):  The confidence level (often 0.90, 0.95, or 0.99)
+            indicates the degree of confidence that the true parameter lies within the interval.
+
+    Returns:
+        (list): returns a list of output statsframe column names
+
+    Equation:
+        .. math::
+
+             \text{CI} = \bar{x} \pm z \left( \frac{\sigma}{\sqrt{n}} \right)
+    """
+    if columns is None or not isinstance(columns, list):
+        raise ValueError("Value passed to 'columns' must be of type list.")
+
+    if not isinstance(confidence_level, float):
+        raise ValueError(
+            r"Value passed to 'confidence_level' must be of type float or int."
+        )
+
+    if confidence_level >= 1 or confidence_level <= 0:
+        raise ValueError(
+            r"Value passed to 'confidence_level' must be in the range of (0, 1)."
+        )
+
+    verify_thicket_structures(thicket.dataframe, columns=columns)
+
+    output_column_names = []
+
+    sample_sizes = []
+
+    # Calculate mean and standard deviation
+    mean_cols = th.stats.mean(thicket, columns=columns)
+    std_cols = th.stats.std(thicket, columns=columns)
+
+    # Convert confidence level to Z score
+    z = scipy.stats.norm.ppf((1 + confidence_level) / 2)
+
+    # Get number of profiles per node
+    idx = pd.IndexSlice
+    for node in thicket.dataframe.index.get_level_values(0).unique().tolist():
+        node_df = thicket.dataframe.loc[idx[node, :]]
+        sample_sizes.append(len(node_df))
+
+    # Calculate confidence interval for every column
+    for i in range(0, len(columns)):
+        x = thicket.statsframe.dataframe[mean_cols[i]]
+        s = thicket.statsframe.dataframe[std_cols[i]]
+
+        c_p = x + (z * (s / np.sqrt(sample_sizes)))
+        c_m = x - (z * (s / np.sqrt(sample_sizes)))
+
+        out = pd.Series(list(zip(c_m, c_p)), index=thicket.statsframe.dataframe.index)
+
+        if thicket.dataframe.columns.nlevels == 1:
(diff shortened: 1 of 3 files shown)


README.md
slabasan, line 1: This update seems unrelated to your confidence interval work, should it go into a separate PR?

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+
+        out = pd.Series(list(zip(c_m, c_p)), index=thicket.statsframe.dataframe.index)
+
+        if thicket.dataframe.columns.nlevels == 1:
+            out_col = f"confidence_interval_{confidence_level}_{columns[i]}"
+        else:
+            out_col = (
+                columns[i][0],
+                f"confidence_interval_{confidence_level}_{columns[i][1]}",
+            )
```
slabasan, line 89: Is it necessary to check if the performance table is has two index levels or 1 to set the column name for the confidence interval? Is it appropriate to check if columns is a list of strings or a list of tuples?

slabasan (approved):

README.md
```diff
@@ -7,8 +7,13 @@
...
 
-A Python-based toolkit for analyzing ensemble performance data. You can find
-detailed documentation, along with tutorials of Thicket in the
+A Python-based toolkit for Exploratory Data Analysis (EDA) of parallel performance data
+that enables performance optimization and understanding of applications’ performance on
+supercomputers. It bridges the performance tool gap between being able to consider only
+a single instance of a simulation run (e.g., single platform, single measurement tool,
+or single scale) and finding actionable insights in multi-dimensional, multi-scale,
+multi-architecture, and multi-tool performance datasets. You can find detailed
+documentation, along with tutorials of Thicket in the
```
ilumsden: I agree with @slabasan's comment. Why is this here?

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns:
+        (list): returns a list of output statsframe column names
+
+    Equation:
+        .. math::
+
+             \text{CI} = \bar{x} \pm z \left( \frac{\sigma}{\sqrt{n}} \right)
+    """
+    if columns is None or isinstance(columns, list) is False:
```
ilumsden: It's more standard to write this if statement this way:
```python
if columns is None or not isinstance(columns, list):
```

ilumsden: In fact, `is False` can lead to unintended consequences. You'd probably be fine here, but you really should only use `is False` or `is True` if you are 100% positive that the return type is a boolean. Some functions return a value that implements `__eq__()` instead of a boolean. In that case, this comparison will never work as intended.

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+        raise ValueError("Value passed to 'columns' must be of type list.")
+
+    if isinstance(confidence_level, (int, float)) is False:
```
ilumsden: Again, use `not isinstance` instead of `is False`

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+        raise ValueError(
+            r"Value passed to 'confidence_level' must be of type float or int."
+        )
+
+    if confidence_level >= 1 or confidence_level <= 0:
```
ilumsden, line 49: Are values of 1 and 0 actually invalid for confidence intervals? I agree with checking for greater than and less than, but I want to make sure we're not excluding valid values here with the "equals" bit.

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+    output_column_names = []
+
+    sample_sizes = []
+
+    # Calculate mean and standard deviation
+    mean_cols = th.stats.mean(thicket, columns=columns)
+    std_cols = th.stats.std(thicket, columns=columns)
+
+    # Convert confidence level to Z score
+    z = stats.norm.ppf((1 + confidence_level) / 2)
```
ilumsden: This doesn't particularly matter, but it would be nice to change `stats` here to something else. Since we refer to `th.stats` quite a bit, just having `stats` might get confusing.

In fact, if this is the only place you're calling into SciPy, just use `import scipy.stats` and change this line to:
```python
z = scipy.stats.norm.ppf(...)
```

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Get number of profiles per node
+    idx = pd.IndexSlice
+    for node in thicket.graph.traverse():
```
ilumsden: It would probably be faster to iterate using the DataFrame than the Graph here. You should be able to do the same with the following for statement:
```python
for node in thicket.dataframe.index.get_level_values(0).unique():
```

If the iteration screws up with that, you can just add `.tolist()` to the end of all of that.

thicket/stats/confidence_interval.py
ilumsden: `columns` should always behave like a list of tuples.

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+        else:
+            out_col = (
+                columns[i][0],
+                f"confidence_interval_{confidence_level}_{columns[i][1]}",
+            )
+
+        output_column_names.append(out_col)
+        thicket.statsframe.dataframe[out_col] = out
+
+    thicket.statsframe.dataframe = thicket.statsframe.dataframe.sort_index(axis=1)
```
ilumsden, line 94: @slabasan do we want the stats functions to be sorting the columns of the statsframe? Personally, I like this, but I'd like to hear your opinion.

thicket/tests/test_stats.py
```diff
@@ -1217,3 +1218,29 @@ def test_cache_decorator(rajaperf_seq_O3_1M_cali):
...
+    ):
+        th.stats.confidence_interval(
+            combined_th, columns=columns, confidence_level="0.95"
+        )
+
+    with pytest.raises(
+        ValueError,
+        match=r"Value passed to 'confidence_level' must be in the range of \(0, 1\).",
+    ):
+        th.stats.confidence_interval(combined_th, columns=columns, confidence_level=95)
```
ilumsden: You should have a test here that actually confirms that the result of the computation is what you expect, at least for a few nodes.

ilumsden (changes requested):
Mostly looks good. Just a few tweaks that I think should be made. The biggest thing is adding a test case that actually confirms that the computations are correct, at least for a few nodes.

thicket/stats/confidence_interval.py
Yejashi: @ilumsden , I think it's useful to have this check.

Confidence intervals are typically set at levels like 90%, 95%, or 99%, which provide a balance between precision and certainty.

Suppose that the confidence level was 100% or 0%:
  * 100%: This would mean that you are certain the true population parameter is within the interval for every possible sample.  To achieve this, the interval would need to be infinitely wide, making it  uninformative because it doesn't give any meaningful estimate of the range in which the true parameter lies.
  * 0%:  This means you have no confidence that the interval contains the true population parameter.  The interval would shrink to a single point.

ilumsden: I get that 1 and 0 aren't super useful. What I want to know is whether they are mathematically valid values. If they are valid, we shouldn't exclude them, which is what this check is doing.

thicket/stats/confidence_interval.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Equation:
+        .. math::
+
+             \text{CI} = \bar{x} \pm z \left( \frac{\sigma}{\sqrt{n}} \right)
+    """
+    if columns is None or not isinstance(columns, list):
+        raise ValueError("Value passed to 'columns' must be of type list.")
+
+    if not isinstance(confidence_level, (int, float)):
```
ilumsden: @Yejashi another thing I just thought of:

This check should only be for `float`. Since the range of this value is (0, 1) (inclusivity TBD), all valid values will be floats, not integers.

thicket/stats/confidence_interval.py
Yejashi: Well it doesn't cause errors but the z-score ends up being infinity if you use a confidence_level of 1 and a 0 if you use a confidence level 0. It is utterly pointless to ever use these parameters in the context of calculating the confidence interval.

I can make it so that they can be accepted as input but again, why?

thicket/stats/confidence_interval.py
slabasan: This will sort the columns alphabetically right? I think the preference is not to re-arrange the columns on the user, so that the new one appears as the far right column. I do like when we re-sort the dataframe if we have a multi-index on the columns so that columns end up under the right header.

Yejashi: @slabasan  Consider this issue #215 . I was always under the impression that we sort the statsframe dataframe for every statsframe function as a matter of principle, but i've become aware that is not the case. It seems Treecs added in this phenomenon over a year ago for all the statsframe operations.

Personally, i think sorting the statsframe dataframe has great utility. It avoids the annoyance of having to scroll back in for an item under the same index but on opposite ends of the dataframe. However, if this is not something we want to force upon the user, then i would at least like/recommend a new function for sorting the statsframe dataframe.

slabasan: For the confidence interval, is there an option to have a "confidence interval" index on the columns with columns underneath? From what I recall, the confidence interval was being added as a column per column and interval.

Yes agree, if we having the column indices appear at both ends of the statsframe, we should sort by default so the statsframe has 1 instance of the column index. But if we have no column index levels, I'm still reluctant to move things on the user.

slabasan: For now, we will sort the columns as part of the confidence interval API to be consistent with other statsframe functions. We will have a discussion about how we want to handle this going forward.

slabasan (approved):

PR #219, opened 2024-10-31
Call squash in groupby

michaelmckinsey1:
Add a conditional check to `groupby` for calling `squash` to update the graph, if nodes disappear from the performance data. This is necessary sometimes for Thickets where `fill_perfdata=False`, as the groupby splits up the profiles into different Thickets, which may contain different nodes. We do the same check in `filter_metadata` and `filter_profile`.


dyokelson (approved):
LGTM

slabasan (approved):

PR #225, opened 2024-11-12
Unicode Histogram in Tree

michaelmckinsey1:
# Summary
Visualize (min, max) values for each node and histogram of profile values using unicode bars. Designed to give user a quick view of data distribution. Number of bars is the square root of the number of profiles for that node.

# Examples
Example for 70 profiles. Notice slight difference vs Matplotlib histogram, since we are limited to 7 bar sizes.

Tree histogram:
image
Equivalent matplotlib histogram:
image

### DIFF
--- a/thicket/thicket.py
+++ b/thicket/thicket.py
@@ -1021,7 +1021,7 @@ class Thicket(GraphFrame):
             min_value (int, optional): Overwrites the min value for the coloring legend. Defaults to None.
             max_value (int, optional): Overwrites the max value for the coloring legend. Defaults to None.
             indices(tuple, list, optional): Index/indices to display on the DataFrame. Defaults to None.
-            histogram (bool, optional): Whether to show a histogram next to each node of the data for all profiles. Defaults to True.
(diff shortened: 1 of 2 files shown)


michaelmckinsey1:
~Should this be on or off by default?~ The default is off by default.

thicket/thicket.py
```diff
@@ -1021,6 +1021,7 @@ def tree(
             min_value (int, optional): Overwrites the min value for the coloring legend. Defaults to None.
             max_value (int, optional): Overwrites the max value for the coloring legend. Defaults to None.
             indices(tuple, list, optional): Index/indices to display on the DataFrame. Defaults to None.
+            histogram (bool, optional): Whether to show a histogram next to each node of the data for all profiles. Defaults to True.
```
slabasan: Looks like the default value is False, can we update these to match? Whether true or false by default?

slabasan (changes requested):

slabasan (approved):

PR #226, opened 2024-11-12
Only Print Duplicates in Error Message

michaelmckinsey1:
From the error message in develop, it is hard to tell which index is the duplicate when there are many profiles. In this example the duplicate is `3723623787`.

develop
Screenshot from 2024-11-11 20-53-17

This PR
Screenshot from 2024-11-11 20-54-44

### DIFF
--- a/thicket/utils.py
+++ b/thicket/utils.py
@@ -75,7 +75,9 @@ def validate_dataframe(df):
             counts = Counter(inner_idx_values)
             duplicates = [item for item, count in counts.items() if count > 1]
             if len(duplicates) > 0:
-                raise DuplicateIndexError(f"Duplicate indices found in DataFrame index.\n\t{duplicates}")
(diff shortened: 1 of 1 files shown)


slabasan:
@michaelmckinsey1 Are you able to amend this PR to check for duplicates earlier? Can we update the error message to report collisions to us?

slabasan (approved):
I can't recall my note about checking for this "earlier", so merging this now and will revisit if it comes up again.

michaelmckinsey1:
> I can't recall my note about checking for this "earlier", so merging this now and will revisit if it comes up again.

I think we only discussed it verbally, but for the future: from what I remember I would basically need to restructure the validation checks to check earlier and add an additional check for hash collisions, which is on my todo list. This PR is strictly improving the way the existing check works.

PR #235, opened 2025-01-22
Add profile_idx_name as member to Thicket class

### LINKED ISSUE #232: Replace hardcoded logic for the "profile" index level with a member variable
Right now, a lot of the logic around the "profile" index level is hardcoded based on the "profile" name. Instead, the logic should be based around a member variable, such as `Thicket.profile_idx_name`. This will give us more flexibility to handle alternate names/name changes in the future. It will also allow us to handle replacing the "profile" index level with another.

michaelmckinsey1:
Resolves #232. Enables replacing the "profile" index.


ilumsden (approved):
LGTM

slabasan (approved):

PR #237, opened 2025-01-23
Add Function to Replace Profile Index

### LINKED ISSUE #233: Add a function to replace the "profile" index level with a metadata column
When dealing with certain datasets (e.g, a collection of profiles per MPI rank), it can be useful to replace the "profile" index level with data from the metadata table (e.g., MPI rank). Right now, we can sort-of do this with `GroupBy.agg(...)`, but it would be nicer to have dedicated functionality for this.

A new function for this would behave along these lines:
1. Take the name of a metadata column as input
2. Check if all the values in the metadata column are unique (equivalent to checking if every profile has an unique metadata value)
3. Copy the metadata column into the performance `DataFrame`
4. Replace the "profile" index level with the copied metadata column
5. Update the row index in the metadata table

michaelmckinsey1:
Resolves https://github.com/LLNL/thicket/issues/233.

### DIFF
--- a/thicket/tests/test_thicket.py
+++ b/thicket/tests/test_thicket.py
@@ -230,3 +230,53 @@ def test_different_profile_idx_name():
         profile_idx_name="profile2",
     )
     assert th.profile_idx_name == "profile2"
+
+
+def test_replace_profile_index(mpi_scaling_cali, intersection, fill_perfdata):
+    th = Thicket.from_caliperreader(
+        mpi_scaling_cali[-1],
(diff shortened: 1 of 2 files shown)


michaelmckinsey1:
depends on #235

michaelmckinsey1:
~Need to add check that new column is unique~ 
8cbb70a

slabasan (approved):

ilumsden (approved):
LGTM

slabasan (approved):

PR #269, opened 2026-01-05
Set Node Ordering `False` by default in Thicket

michaelmckinsey1:
This is `True` by default in Hatchet, but for Thicket, set `node_ordering=False` to match user expectations.


thicket/tests/conftest.py
```diff
@@ -79,14 +79,14 @@ def stats_thicket_axis_columns(
     """
     th_cuda128_1 = Thicket.from_caliperreader(
         rajaperf_cuda_block128_1M_cali[0:4],
-        node_ordering=True,
+        node_ordering=False,
```
michaelmckinsey1: This should fail CI, but the test needs to be fixed

